PR #5, opened 2026-03-21
fix: security fixes — AppleScript injection, path traversal, batch DoS, PII in logs + dep bumps

chrischall:
## Summary

- **AppleScript injection (Critical):** Add `MESSAGE_ID_SCHEMA` (`/^\d+$/`) to all 11 message tool `id` schemas; add `DATE_FILTER_SCHEMA` allowlist + `escapeForAppleScript()` for `dateFrom`/`dateTo` in `search-messages`
- **Path traversal (High):** Validate `savePath` is within `homedir()/tmp/private/tmp/Volumes` and `attachmentName` contains no `/`, `..`, or null bytes in `saveAttachment()`
- **Batch DoS (High):** Cap all 6 batch operation `ids` arrays at 100 with `BATCH_IDS_SCHEMA`
- **PII leakage in debug logs (Medium):** Redact AppleScript double-quoted string content from `scriptPreview` before logging
- **Deps:** Bump `@modelcontextprotocol/sdk` 1.4.1→1.27.1, `@types/node` ^20→^25, `vitest`/`@vitest/coverage-v8` ^2→^4, `eslint` ^9→^10 (clears all 7 npm audit vulnerabilities)

## Test Plan

- [x] `npm test` — 56 tests pass (3 test files: `src/index.test.ts`, `src/services/appleMailManager.test.ts`, extended `src/utils/applescript.test.ts`)
- [x] `npm run typecheck` — no errors
- [x] `npm run lint` — no errors
- [x] `npm run build` — clean build

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "apple-mail-mcp": "build/index.js"
       },
       "devDependencies": {
+        "@eslint/js": "^10.0.1",
         "@types/node": "^25.5.0",
         "@typescript-eslint/eslint-plugin": "^8.0.0",
         "@typescript-eslint/parser": "^8.0.0",
@@ -200,6 +201,27 @@
       },
       "engines": {
         "node": "^20.19.0 || ^22.13.0 || >=24"
+      }
+    },
+    "node_modules/@eslint/js": {
+      "version": "10.0.1",
+      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-10.0.1.tgz",
+      "integrity": "sha512-zeR9k5pd4gxjZ0abRoIaxdc7I3nDktoXZk2qOv9gCNWx3mVwEn32VRhyLaRsDiJjTs0xq/T8mfPtyuXu7GWBcA==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "^20.19.0 || ^22.13.0 || >=24"
+      },
+      "funding": {
+        "url": "https://eslint.org/donate"
+      },
+      "peerDependencies": {
+        "eslint": "^10.0.0"
+      },
+      "peerDependenciesMeta": {
+        "eslint": {
+          "optional": true
+        }
       }
     },
     "node_modules/@eslint/object-schema": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,18 +12,18 @@
         "darwin"
       ],
       "dependencies": {
-        "@modelcontextprotocol/sdk": "1.4.1",
+        "@modelcontextprotocol/sdk": "^1.27.1",
         "zod": "^3.22.4"
       },
       "bin": {
         "apple-mail-mcp": "build/index.js"
       },
       "devDependencies": {
-        "@types/node": "^20.0.0",
+        "@types/node": "^25.5.0",
         "@typescript-eslint/eslint-plugin": "^8.0.0",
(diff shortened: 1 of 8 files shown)


sweetrb:
Hey @chrischall — thanks for putting this together! Really appreciate you taking the time to identify and fix these security gaps. The AppleScript injection and path traversal issues especially are the kind of thing that's easy to overlook in an MCP project like this.

I've gone through the code and overall it's solid work. A few things I'd like to discuss before merging:

**Symlink bypass on path traversal check (important)**

The `savePath.startsWith(prefix)` check in `saveAttachment()` validates the string representation of the path, but doesn't resolve symlinks or `..` segments first. Something like `/tmp/../../etc/passwd` would pass the prefix check since it starts with `/tmp`. Adding a `path.resolve(savePath)` (or `fs.realpathSync()`) before the prefix comparison would close that gap.

**Unquoted ID interpolation in AppleScript**

In `saveAttachment()`, the message ID gets interpolated directly into the AppleScript as `whose id is ${id}` without quotes. The Zod schema keeps this safe right now since it enforces numeric-only, but it's a bit fragile — if someone touches that schema later they might not realize this is a raw interpolation. Worth either quoting it (`whose id is "${id}"`) or casting with `Number(id)` as a defense-in-depth measure.

**Consider splitting the dependency bumps into a separate PR**

The security fixes are focused and clean, but the `@modelcontextprotocol/sdk` jump from 1.4.1 to 1.27.1 is a pretty big leap that brings in Express 5.x, Hono, JWT libs, and a bunch of new transitive deps. Same with ESLint 9 to 10 (breaking changes in config format). Bundling these with security fixes makes it harder to isolate if something breaks. Would you be open to splitting those into a follow-up PR?

**Minor stuff / nice-to-haves**

- The `escapeForAppleScript()` call on `dateFrom`/`dateTo` in `searchMessages()` is technically a no-op since the Zod schema already blocks all the characters it would escape. Not a problem (belt-and-suspenders is fine), but a quick comment noting that would help future readers.
- No tests for the `searchMessages` date filter interpolation path itself — the schema validation is well-tested, but an integration-level test showing a valid date string flows correctly into the AppleScript would round things out.
- `attachmentName` validation blocks `/`, `..`, and null bytes but not `\`. Probably fine on macOS but worth a quick mention in the comments for completeness.

None of this is meant to diminish the work here — the security findings are legit, the fixes are well-targeted, and the test coverage is thorough. Really nice to see the regression test for template IDs too (making sure `tmpl_N` format didn't get caught by the numeric regex).

Thanks again for the contribution — looking forward to getting this merged once we sort out the path resolution piece!

sweetrb:
Hey @chrischall — thank you for the thorough security audit and the PR! The findings were spot-on and really valuable.

After reviewing, I decided to implement the fixes on a separate branch ([#6](../pull/6)) so I could:

1. **Apply the review feedback directly** — `path.resolve()` before prefix check, `Number(id)` defense-in-depth, backslash blocking in attachment names
2. **Split out the dependency bumps** — the SDK 1.4.1→1.27.1 and ESLint 9→10 jumps bring significant transitive dep changes that I want to evaluate separately
3. **Add comprehensive tests** — 34 new unit tests + 44 integration tests against real Mail.app

All six of your security findings are addressed in #6. Your contribution made this possible — really appreciate the time you put in. 🙏

PR #8, opened 2026-04-20
Fix attachment handling with MIME source fallback

kevinmay-scoutsolutions:
## Summary

AppleScript's `mail attachments` object returns empty collections across **all account types** (iCloud, Google, Exchange) when attachments are embedded as MIME parts in the message source. `list-attachments` and `save-attachment` both fail to see attachments that are clearly present in the email.

This PR adds a MIME source fallback: when AppleScript returns empty, parse the raw `source of msg` for `Content-Disposition: attachment` blocks and surface those instead.

## Problem

I spent a full inbox processing session confirming this wasn't a configuration issue. `list-attachments` returned "No attachments found" on:
- A test email with a PDF attachment sent to my iCloud account (ID confirmed in Apple Mail UI, attachment visible)
- Real forwarded emails in Exchange with PDF/DOCX attachments
- Emails in Google Workspace account

In all cases, the MIME source contained proper `Content-Type: application/pdf` and `Content-Disposition: attachment; filename="..."` headers with base64-encoded bodies, but AppleScript's `mail attachments of msg` returned an empty collection.

This is a limitation of how Mail.app's AppleScript bridge exposes attachments - MIME-encoded attachments aren't surfaced through the attachment object model, even though they're visible in the UI.

## Solution

Two-attempt pattern. AppleScript is tried first (works for some configurations), then falls back to MIME parsing:

1. **`listAttachments`**: If AppleScript returns empty, call `getRawSource(id)` and parse MIME `Content-Disposition: attachment` blocks for filename/size/type metadata.

2. **`saveAttachment`**: If AppleScript can't find the attachment, extract the base64 from MIME source, decode, and write to disk. Preserves all existing path validation (no traversal, allowed dirs only).

3. **`hasAttachments`**: Previously hardcoded to `false`. Now checks AppleScript count first, falls back to checking for `Content-Disposition: attachment` in the MIME source.

## New files

- `src/utils/mimeParse.ts` - Standalone MIME parser (no dependencies) with `parseMimeAttachments()` and `extractMimeAttachment()` functions
- `src/utils/mimeParse.test.ts` - 10 unit tests covering single/multi attachments, inline dispositions, empty input, size estimation

## Test plan

- [x] `npm test` - 73 tests passing (10 new + 63 existing), zero regressions
- [x] `npm run typecheck` - clean
- [x] `npm run build` - clean
- [x] Verified `list-attachments` now returns attachments on iCloud account (test PDF)
- [x] Verified `list-attachments` returns attachments on Exchange account (Ceragon/Outlook)
- [x] Verified `save-attachment` writes valid files to disk via MIME fallback
- [x] Verified path traversal protection still enforced

## Notes

The two-attempt pattern means this fix degrades gracefully. If Apple ever fixes the AppleScript object model, the original code path handles it. If not, we fall through to MIME parsing. No hard dependency on either path working alone.

The MIME parser itself is generic RFC 2045/2046 - could potentially be extracted as a separate utility module if useful elsewhere.

### DIFF
--- a/src/services/appleMailManager.ts
+++ b/src/services/appleMailManager.ts
@@ -14,10 +14,11 @@
  */
 
 import { spawnSync } from "child_process";
-import { existsSync } from "fs";
+import { existsSync, writeFileSync } from "fs";
 import { isAbsolute, resolve } from "path";
 import { homedir } from "os";
 import { executeAppleScript } from "@/utils/applescript.js";
+import { parseMimeAttachments, extractMimeAttachment } from "@/utils/mimeParse.js";
 import type {
   Message,
   MessageContent,
@@ -493,7 +494,18 @@ export class AppleMailManager {
                 set msgDeleted to deleted status of msg as string
                 set msgMailbox to name of mb
                 set msgAccount to name of acct
-                return msgSubject & "|||" & msgSender & "|||" & msgDate & "|||" & msgRead & "|||" & msgFlagged & "|||" & msgJunk & "|||" & msgDeleted & "|||" & msgMailbox & "|||" & msgAccount
+                set hasAtt to "false"
+                try
+                  set attCount to count of mail attachments of msg
+                  if attCount > 0 then set hasAtt to "true"
+                end try
+                if hasAtt is "false" then
+                  try
+                    set rawSrc to source of msg
+                    if rawSrc contains "Content-Disposition: attachment" then set hasAtt to "true"
+                  end try
+                end if
+                return msgSubject & "|||" & msgSender & "|||" & msgDate & "|||" & msgRead & "|||" & msgFlagged & "|||" & msgJunk & "|||" & msgDeleted & "|||" & msgMailbox & "|||" & msgAccount & "|||" & hasAtt
               end if
             end try
           end repeat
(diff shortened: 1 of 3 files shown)


sweetrb:
Thanks for the thorough investigation and the clean two-attempt fix, @kevinmay-scoutsolutions! Shipped in [v1.5.0](https://github.com/sweetrb/apple-mail-mcp/releases/tag/v1.5.0) along with a few follow-ups stacked on top of your commit:

- Extended `hasAttachments` to `list-messages` (fast AppleScript-count path only, to keep list perf — details in the commit message)
- Recursive descent into nested `multipart/alternative` and `multipart/related` so attachments alongside HTML bodies or inline images are found
- Added quoted-printable / 7bit / 8bit transfer encoding support to the MIME fallback
- Inline comments on the `getMessage` source-scan perf tradeoff and the 120s `getRawSource` timeout

Your MIME parser is the core of all of it — great work. Published to npm as `apple-mail-mcp@1.5.0`.

kevinmay-scoutsolutions:
@sweetrb thanks for running with this... really appreciate the extensions. The nested multipart descent and the encoding support close out gaps I knew about but didn't have runway to chase. Smart call keeping hasAttachments on the fast AppleScript count for list-messages too — nice perf instinct. Glad it landed clean, and thanks for the kind words :)

PR #13, opened 2026-05-27
fix(search): wire from/subject/isRead/isFlagged filters into search-messages

kevinmay-scoutsolutions:
## Summary

`search-messages` declares `from`, `subject`, `isRead`, and `isFlagged` in its input schema, but the handler only forwarded `query/mailbox/account/limit/dateFrom/dateTo` to `searchMessages()`. The four filters were destructured but never passed through, so they silently did nothing — e.g. `from="bob@x.com"` returned the same results as no filter at all.

## Problem

Because the params are *declared* in the schema, callers (and LLM agents) reasonably assume they work. The failure is silent: no error, just unfiltered results. I confirmed it empirically — four different `from` values against the same mailbox returned byte-identical result sets. The underlying `searchMessages()` only ever built `whose subject contains X or sender contains X` from `query`; the other four were dead parameters.

## Fix

- Forward `from`, `subject`, `isRead`, `isFlagged` from the handler into `searchMessages()`.
- Build them into the AppleScript `whose` clause as AND filters. The existing `query` (subject-OR-sender) is now parenthesized so precedence holds when combined: `(subject contains X or sender contains X) and ...`.
- Extract the clause logic into a pure, exported `buildSearchCondition()` with unit tests (9), covering the `isRead:false` / `isFlagged:false` not-dropped regression (a naive `if (isRead)` would reintroduce it), OR-grouping, and quote/backslash escaping.
- Clarify in the schema that `from`/`subject` are substring matches (not exact-address).

## Notes

- `read status is true/false` and `flagged status is false` are valid Mail.app `whose` filters. `from` maps to `sender contains`, a substring match over the full `Display Name <addr>` string (documented accordingly).
- Date filtering is unchanged (stays post-fetch) — `whose` date comparisons are unreliable in Mail.app.
- No change to `list-messages`, which already wired `from` through separately.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` — 90 passing (9 new)
- [x] `npm run build`
- [x] Live check against Mail.app: `from="<addr>"` returned only that sender vs a mixed unfiltered baseline; `isFlagged=true` returned only flagged messages — confirms the generated `whose` clauses execute and filter correctly end-to-end.


sweetrb:
Thanks for this, @kevinmay-scoutsolutions — really nice work.

A few things I want to call out that I appreciated:

- **The diagnosis.** Catching that the four params were declared in the schema but never destructured out of `withErrorHandling`'s callback is exactly the kind of bug that's easy to miss in review and impossible to detect from the outside — it just looks like search is working, the filters just don't narrow.
- **`typeof === "boolean"` instead of `if (isRead)`.** This is the kind of subtle thing that distinguishes a real fix from a fix that introduces its own bug. The dedicated regression test for the `false` case is a chef's kiss.
- **Extracting `buildSearchCondition()` as a pure function.** Decouples the actual clause-building from Mail.app, which is the only piece of this whole module that's testable without a live mailbox. Future filter additions will be much easier to verify.
- **Parenthesizing the existing `query` OR-clause.** Easy to overlook when adding ANDs and it would have silently flipped precedence on combined filters.

Verified locally against my real inbox: `from`, `subject`, `isRead=false`, and `isFlagged=true` all narrow correctly end-to-end. Merged and released as **v1.5.2** — CI is auto-publishing to npm now.

Thanks again, and please keep the contributions coming.

PR #14, opened 2026-05-27
fix(move): support nested destination mailboxes in move-message

kevinmay-scoutsolutions:
## Summary

`move-message` / `batch-move-messages` could not move a message into a **nested mailbox**. The destination was resolved as `mailbox "X" of account "Y"`, which only finds certain top-level mailboxes, so moving to e.g. a "Moore" subfolder failed outright.

## Problem

`mailboxes of account` already returns a **flat** list that includes nested mailboxes (named by path, e.g. `Processed/Vendors`). Two gotchas:
- Descending into them via `mailboxes of mb` is unreliable — it double-prepends the parent path (`Processed/Processed/...`) and throws `-1728`.
- Addressing `mailbox "X" of account "Y"` fails for these server mailboxes even when `X` appears in the flat list.

## Fix

- Match the destination against the flat `mailboxes of account` list by exact name and use the **reference directly** (no descent, no by-name re-addressing).
- **Account-scoped + ambiguity-aware:** if the name matches more than one mailbox in the account, return an error instead of guessing — silently moving mail to the wrong folder is worse than failing. (Real case: an Exchange account with two `Ceragon` mailboxes.)
- The flat list already covers nested mailboxes, so the source-message lookup reaches messages in subfolders too.
- Distinct errors (destination not found / ambiguous / message not found) now propagate through `batch-move-messages` results via an internal `moveMessageInternal`.

## Notes

- No new unit test: destination resolution runs in AppleScript against the live Mail object model, not a pure TS function. Validated via live integration test instead.

## Test plan

- [x] `npm run typecheck` / `npm run lint` / `npm run build`
- [x] `npm test` — 81 passing
- [x] committed `build/` matches source
- [x] Live (Exchange account):
  - move to nested `Moore` subfolder → **succeeds**, message lands in Moore
  - move to `Ceragon` (two mailboxes share the name) → **fails** with `ambiguous (2 matches)`, message not moved
  - move to a nonexistent mailbox → **fails** with `not found`


sweetrb:
Thanks again, @kevinmay-scoutsolutions — second nice catch in a row.

A few things from my review:

- **The diagnosis is the kind that's only obvious once someone's written it down.** `mailboxes of account` returning a flat path-named list while `mailbox "X" of account "Y"` only resolves top-level — that's a real Mail.app quirk, and you not only found it but explained exactly why descending via `mailboxes of mb` would have looked like a tempting fix and is in fact a trap (the double-path-prepend / `-1728`).
- **Ambiguity-aware refusal is the right default.** When I ran your code against my Gmail account, it correctly flagged that I have two `Drafts` mailboxes and refused to guess. Then I ran the *old* code on the same input and it silently moved the message to one of them — no way to tell which. So this PR actually closes a quiet data-safety footgun on top of fixing the stated bug. I called that out in the CHANGELOG; you undersold it in your description.
- **`moveMessageInternal` returning structured errors so `batch-move-messages` can surface them.** Strict improvement to observability — the previous "Failed to move message" string was useless for diagnosing batch failures. The boolean-returning `moveMessage()` wrapper keeps the public API compatible. Clean refactor.
- **Honest call on testing.** "No unit test because resolution runs in AppleScript against the live Mail object model" is the correct trade-off, and the live integration cases you listed (Exchange `Moore` success, Gmail-style ambiguity, nonexistent destination) are exactly the three branches that needed validation. I reproduced your error-path results on my end (one nested-mailbox setup, one ambiguous setup) and they came out as advertised.

There was a trivial merge conflict in the generated sourcemap from #13 landing first — resolved by regenerating from source. Merged and released as **v1.5.3**.

Really appreciate the contributions. If you've got more in the queue, send them along.

Issue #43, opened 2026-06-18
Optional IMAP backend for read/search/manage on IMAP accounts (fast server-side search; fixes Gmail slowness + server-mailbox ops)


sweetrb:
## Summary

Add an **optional, opt-in IMAP backend** for read/search/manage operations on IMAP-capable accounts (Gmail/Workspace, iCloud, generic IMAP), mirroring the `transport:"smtp"` pattern from #12. AppleScript-over-Mail.app stays the **default**; IMAP is used only when an account is configured for it. This turns the two worst pain points — slow/false-empty search (#24/#16) and un-deletable/un-renamable server mailboxes (#42) — into fast, correct, server-side operations, without giving up the zero-config AppleScript path for everything else.

## Motivation — measured

Live benchmark on a real Google Workspace account (2026-06-18), reusing an existing Keychain app password:

| Operation | AppleScript (current) | IMAP (server-side) |
|---|---|---|
| SEARCH `SUBJECT "the"` in **All Mail** | **times out ~55s, returns a FALSE empty** | **355 hits in 0.38s** |
| SEARCH `SINCE 01-Jun-2026` | (same timeout class) | 214 hits in 0.56s |
| LIST folders | fast | 0.34s |
| Login + list + 2 searches + header fetch | — | **3.7s total** |

~150× faster **and correct**, on the exact mailbox where AppleScript can't complete. Cause is structural: Gmail exposes labels as IMAP folders where All Mail holds every message; AppleScript `whose`/iteration evaluates the predicate client-side across the whole mailbox through the slow Apple Event bridge, while IMAP `SEARCH` runs on Google's servers.

## Why SMTP isn't enough / why IMAP

- **SMTP** (already used, #12) is send-only — no read/search/list/move/delete.
- **IMAP** is the read/manage protocol Gmail's own Mail GUI effectively uses; it gives server-side SEARCH (fixes slowness) and proper folder/label + message ops (fixes the #42 server-mailbox delete/rename gap).
- **Gmail API** (REST/OAuth) is richer still (Gmail query syntax, threads, batch, push) but heavier (Google Cloud project + OAuth) and Gmail-only — out of scope here; IMAP covers the need with far less setup.

## Design (hybrid, opt-in)

[...]

Issue #42, opened 2026-06-18
delete-mailbox/rename-mailbox/draft-delete should fail with a clear 'not supported for server mailboxes' error


sweetrb:
## Summary

`delete-mailbox`, `rename-mailbox`, and draft deletion/move return a **generic failure** when the target lives on a server-side account (IMAP/Gmail-Workspace/iCloud/Exchange), where Mail.app's AppleScript handler simply cannot perform the operation. Callers (and LLM agents) get an unhelpful `Failed to delete mailbox "X"` / `Failed to delete message "N"` with no indication that this is an **unsupported-on-this-account-type** condition rather than a transient error worth retrying.

Proposal: detect this case and return an **actionable, specific** error — e.g. *"Mail.app cannot delete server-side (IMAP/Gmail/iCloud/Exchange) mailboxes via AppleScript; delete it in Mail directly."* — instead of the generic failure.

## Observed (build 1.6.9, 2026-06-18, Google Workspace account)

A reversible smoke test on a personal Google Workspace account:

- `create-mailbox` → **succeeds** (AppleScript implements this handler).
- `rename-mailbox` (create-target + move + delete-source) → **fails**; the target got created (orphan empty mailbox left behind) then the op aborted. The #33 safety held — source mailbox + its message were fully preserved — but the error returned was just `Failed to rename mailbox`.
- `delete-mailbox` → **fails** with `Failed to delete mailbox "..."`.
- Deleting a **draft** via `delete-message`, and moving it via `move-message` → both **fail** with generic messages.

So the failure is **verb- and account-type-specific**, not a permissions or connectivity problem: creation works in the same session; deletion/rename of server mailboxes and draft mutation do not.

## Root cause (platform limitation, not a code defect)

The Mail **GUI** and the **AppleScript** interface are different entry points. The GUI runs Mail's full, maintained mailbox-management logic (and the correct Gmail-label / IMAP folder semantics). The AppleScript `delete`/`rename mailbox` handlers are an older, incomplete implementation that throws `AppleEvent handler failed` for server-side mailboxes (and for draft mutation). This mirrors the platform note already in `docs/STABILITY-PERF-AUDIT-2026-06-17.md` ("Mail.app cannot delete iCloud mailboxes via AppleScript") — it's the same class of limitation, broader than just iCloud (Gmail/Workspace folders and drafts too). AppleScript mailbox delete generally only works for local "On My Mac" mailboxes.

## Proposed behavior

[...]

PR #44, opened 2026-06-18
feat(imap): Phase-1 IMAP read backend for search/list (#43, PoC)

sweetrb:
**Draft / proof-of-concept for #43 — not for merge as-is.**

## What

Opt-in IMAP backend for the two read paths that suffer most from AppleScript on Gmail. When an account is configured for IMAP, `search-messages` and `list-messages` run a **server-side IMAP search** (via `imapflow`) instead of AppleScript. Everything else — and any account *not* configured for IMAP — is unchanged.

## Config (mirrors the transport:smtp pattern)

```
APPLE_MAIL_MCP_IMAP_USER=rob@example.com      # required; enables IMAP + login address
APPLE_MAIL_MCP_IMAP_ACCOUNT=rob@example.com   # Mail account name to match for routing (default = USER)
APPLE_MAIL_MCP_IMAP_HOST=imap.gmail.com       # default
APPLE_MAIL_MCP_IMAP_PORT=993                  # default, implicit TLS
# password: APPLE_MAIL_MCP_IMAP_PASSWORD, or Keychain via
APPLE_MAIL_MCP_IMAP_KEYCHAIN_SERVICE=...  APPLE_MAIL_MCP_IMAP_KEYCHAIN_ACCOUNT=...
```

Routing is conservative: only calls whose explicit `account` matches the configured IMAP account go to IMAP; everything else falls through to AppleScript.

## Result (live, Google Workspace account)

| Call | AppleScript today | This PoC (IMAP) |
|---|---|---|
| `search-messages` query in All Mail | ~55s timeout → **false empty** | **5 of 434 matched, ~5s, correct** |
| `list-messages` INBOX | empty / times out | 2 real messages, ~5s |

## Scope / not-yet

- Read-only: `search-messages`, `list-messages`. `get-message` + all mutations stay on AppleScript (results note that IDs are IMAP UIDs).
- No connection pooling yet (each call connects/​logs-out — ~5s incl. handshake; pooling would cut this).
- Gmail mailbox-name mapping is a small static table; a fuller version would resolve via LIST/special-use flags.
- Phase 2 (mutations + folder ops via IMAP, which would also address the IMAP slice of #42) is out of scope here.

11 new unit tests (injected mock client, no network); 149 tests pass; typecheck + lint clean.

Refs #43.


sweetrb:
Productionized for merge: rebased on main (through 1.6.10), added the README IMAP backend section + CHANGELOG entry. Scope unchanged from the PoC — opt-in, read-only Phase 1 (search/list), additive with zero regression when IMAP isn't configured. 152 tests (12 IMAP, mocked) + lint + typecheck + build green. Phase 2 (IMAP mutations/folder ops) tracked as future work under #43/#42.

PR #46, opened 2026-06-18
feat(imap): Phase-2 folder ops via IMAP (create/rename/delete-mailbox) (#43)

sweetrb:
Phase 2 of #43 (builds on the Phase-1 read backend in #44 / v1.7.0).

When an account is IMAP-configured, `create-mailbox`, `rename-mailbox`, and `delete-mailbox` route to IMAP (`imapflow` `mailboxCreate`/`mailboxRename`/`mailboxDelete`) instead of AppleScript. IMAP's `CREATE`/`RENAME`/`DELETE` operate on the real server folder hierarchy, so they succeed on exactly the **iCloud / Gmail / Workspace / Exchange** mailboxes where Mail.app's AppleScript bridge throws `AppleEvent handler failed` — closing the IMAP slice of #42.

- Delete/rename resolve the target by `list()` (full-path then leaf-name, case-insensitive) and fail clearly with "not found" **without acting** when the mailbox is absent.
- Opt-in/additive: accounts without IMAP configured use AppleScript (and get the #42 actionable error). Routing only fires when the explicit `account` matches the configured IMAP account.
- Message-level mutations (mark/flag/move/delete-message) stay on AppleScript pending a UID-aware design — the IMAP read path reports per-mailbox UIDs, a different namespace from Mail message ids, so auto-routing them by id would risk acting on the wrong message.

**Verification:** 157 unit tests (5 new folder-op tests, mocked client) + lint + typecheck + build green. Live validation against a real iCloud account (deleting the two orphan test mailboxes) is the next step before merge.

Refs #43, #42.


sweetrb:
Live-validated against a real iCloud account (imap.mail.me.com, app-specific password in Keychain): `imapDeleteMailbox` deleted two server-side iCloud mailboxes that AppleScript could not touch (`AppleEvent handler failed`), and a follow-up IMAP `list()` confirmed they're gone server-side. This is exactly the #42 case the AppleScript path can only fail on. (Mail.app's local sidebar cache lags the server deletion — the documented IMAP-vs-AppleScript divergence — but the server is authoritative and clean.)

PR #65, opened 2026-06-26
fix(mailbox): guard structural ops on disabled accounts; roll back orphaned rename destination

sweetrb:
## Why

While cleaning up two leftover empty iCloud mailboxes (`_amcp_rename_test_src` / `_amcp_rename_test_dst`), we traced them to two real gaps in the AppleScript mailbox ops:

1. **`renameMailbox` leaves an orphan on move failure.** Rename is emulated as create-new + move-all + delete-old. If the move step errors, the freshly-created (empty) destination was left behind — the origin of those stray test mailboxes.
2. **Structural ops against a *disabled* account fail opaquely.** A disabled account has no live server session, so `create`/`delete`/`rename` fail inside Mail with a bare `AppleEvent -10000`. Worse, `describeMailboxOpError` misattributed that -10000 to the "#42 server-side-mailbox limitation," and a half-built rename could leave a mailbox behind.

## What

- **Disabled-account guard.** `createMailbox` / `deleteMailbox` / `renameMailbox` now probe the account's live `enabled` state (fresh, bypassing the 60 s cache) and refuse up front with an actionable message instead of attempting the doomed op. **Fails open** — an indeterminate probe never blocks an otherwise-valid op. AppleScript backend only; direct-IMAP accounts talk to the server independent of Mail's enabled toggle and are routed before this.
- **Rename rollback.** On a failed move, the just-created destination is deleted — **only when empty**, so a partial move (some messages already relocated) still keeps both mailboxes intact and never destroys mail. The error now states whether the destination was rolled back or left for manual cleanup.
- **`createMailbox()` now returns `{ success, error }`** (was `boolean`) so failure reasons surface through the `create-mailbox` tool. Updated its one caller in `index.ts`.
- **`describeMailboxOpError`** extended to cover the `create` verb.

## Tests

- New `appleMailManager.guard.test.ts` (9 cases, `executeAppleScript` mocked): refuse-on-disabled for all three ops (asserting the create/move/delete script never runs), fail-open on an indeterminate probe, rollback of an empty destination, retention + manual-cleanup message for a non-empty destination, and the new `create` verb mapping.
- Full suite: **265 unit tests pass**, typecheck + lint clean.
- **Verified live against the real disabled iCloud account**: `create`/`delete`/`rename` all refused with the guard message and **no orphan mailbox was created** (mailbox list unchanged) — i.e. the exact scenario that produced the `_amcp_rename_test_*` ghosts no longer leaves a trace.

## Note

Unreleased — no version bump (that's the manual release step).


sweetrb:
Shipped in **v2.5.0** (commit b50163a on `main`). These exact changes — the disabled-account guard for create/delete/rename plus the empty-orphan `rename-mailbox` rollback, and the guard unit test — were folded into the 2.5.0 release commit alongside the reply/forward-via-SMTP-threading work. The only delta versus this branch is Prettier formatting applied by the pre-commit hook (verified: substance is byte-identical). Closing as shipped and deleting the branch.

PR #73, opened 2026-06-30
[DO NOT MERGE] version-guard self-test

sweetrb:
Throwaway PR: a src change with NO version bump. Expect the version-guard check to FAIL. Will be closed after verification.


sweetrb:
Self-test passed: version-guard correctly failed the un-bumped src change. Closing.

PR #79, opened 2026-07-06
Bundle both entrypoints with esbuild so a bare clone runs without node_modules

### LINKED ISSUE #78: Marketplace clone can't run the server: node_modules missing after re-clone
Companion to sweetrb/apple-notes-mcp#68 (same failure, same night, same mechanism). Claude Code's marketplace auto-update re-clones the plugin from scratch, and the fresh clone has the committed `build/` but no `node_modules/`. Since the plugin launches with `node ${CLAUDE_PLUGIN_ROOT}/build/index.js`, the server dies on its first import at every session start:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@modelcontextprotocol/sdk'
imported from .../build/index.js
```

`pnpm install && pnpm run build` in the clone fixes it until the next marketplace refresh re-clones and wipes it again.

The durable fix is a dependency-free bundled entrypoint (esbuild `--bundle --platform=node`), since the marketplace install path never runs a package manager and a `prepare` script can't fire. Full detail and repro in the apple-notes issue.

Seen on v2.8.2, macOS, Claude Code marketplace install.

oliverames:
Fixes #78. Companion to sweetrb/apple-notes-mcp#69, which is the same change for the notes server.

## What changed

`npm run build` now typechecks with `tsc --noEmit` and then bundles `src/index.ts` and `src/cli.ts` with esbuild into self-contained `build/index.js` and `build/cli.js`. A bare git clone, which is exactly what Claude Code's marketplace auto-update produces, now launches the server with nothing but Node installed, and `apple-mail-send` keeps working as a standalone bin. Only the two bundled entrypoints are tracked in git; the per-module compiled files and `.d.ts` output are removed and stay ignored. `tsc-alias` is dropped because esbuild resolves the tsconfig path aliases natively.

## Why

The committed `build/` gave a fresh clone the entrypoints, but the compiled output still imported its runtime dependencies from `node_modules/`, which a clone never has. The marketplace refresh path never runs a package manager, so every auto-update killed the server at session start with `ERR_MODULE_NOT_FOUND` on `@modelcontextprotocol/sdk`, and the failure was silent unless you went digging in Claude Code's MCP logs.

## Two wrinkles worth knowing about

- Unlike the notes server, this one bundles CJS dependencies (nodemailer, imapflow) that require Node builtins dynamically, which esbuild's ESM output rejects at runtime with `Dynamic require of "events" is not supported`. The build therefore adds a `createRequire` banner. esbuild hoists the entry shebang above the banner and renames the module's own `require` declaration, so both bins stay intact; I verified all of that by running the bundles, not just building them.
- `src/index.ts` reads `../package.json` at runtime for its version string, so the artifact is two files plus `package.json` rather than strictly standalone. I left that alone: every distribution layout ships `package.json` in that spot, and keeping `src/` untouched keeps this PR outside version-guard's shipped-code rule.

## What the `.d.ts` removal means

The old build emitted declarations and `package.json` pointed `types` at `build/index.d.ts`, so this PR does shrink the published npm surface: no more type declarations. My read is that nothing imports an MCP server binary as a typed library, and the sibling notes repo has shipped without working types all along. If you want to keep shipping declarations, the build can run `tsc --emitDeclarationOnly` alongside esbuild instead; say the word and I'll switch it.

## Verification

- `lint`, `typecheck`, `test` (345 passing across 25 files), `format:check`, and `build` are all green locally.
- Live probe of the real failure mode: fresh `git clone` of this branch into an empty directory, no install step. `initialize` returns `serverInfo` for apple-mail 2.8.2, `tools/list` returns all 46 tools with nothing on stderr, and `node build/cli.js --help` prints the usage and exits 0.

## Left out of scope

No version bump, since nothing under `src/` changed. The fix matters most for the git and marketplace paths, which track main; whether to cut an npm release for it is your call.


sweetrb:
Merged and released as **v2.8.3**. Sharp root-cause, @oliverames — committing `build/` only ever fixed the missing-entrypoint half, and pinning the failure to the marketplace's re-clone → `node build/index.js` path (no package-manager step anywhere between refresh and start) is exactly it. The `createRequire` banner to keep nodemailer/imapflow's dynamic Node-builtin `require`s working under esbuild's ESM output is a clean touch. Independently verified the committed bundle runs a full MCP `initialize` handshake from a bare-clone layout with **no `node_modules`**. Thanks for the trio of contributions across the two servers today.

PR #93, opened 2026-07-15
fix(get-message): return RFC Message-ID on AppleScript path + scoped fetch for Sent Items

Goaleve1:
## Summary

Fixes two issues with `get-message` that prevented callers from reliably retrieving a message's stable RFC Message-ID and from opening messages in the Sent Items mailbox.

## Root cause 1 — RFC Message-ID returned `null` on the AppleScript path

A prior change added Message-ID retrieval only to the IMAP branch. When the server runs via AppleScript (Mail.app automation), `get-message` still returned `null` for the Message-ID, so callers had no stable `<...@...>` identifier to key off of.

**Fix:** Read Mail.app's native `message id` property on the AppleScript path so `get-message` returns the stable RFC Message-ID there too.

## Root cause 2 — Sent Items messages resolved as "message not found"

Resolving a message by id performed an unscoped scan across the entire message store (hundreds of mailboxes), which timed out for messages living in Sent Items. The message was present but never reached before the operation gave up.

**Fix:** Build an `id -> { account, mailbox }` index and perform a scoped, single-mailbox fetch. The message body resolves directly without a full-store scan.

## Result

A single `get-message` call now returns **both** the message body **and** the stable RFC Message-ID, including for messages in Sent Items.

## Testing

- `tsc --noEmit` typecheck: clean
- Test suite: **368 tests pass** (26 files), including added coverage for the MIME/Message-ID parsing changes.


sweetrb (changes requested):
This is an outstanding first contribution — thank you. I did a full review + local verification before writing this:

**Verified on my side:**
- ✅ Rebuilt `build/index.js` from your source: **byte-identical** to your committed bundle.
- ✅ All gates green locally (lint / format:check / typecheck / full test suite incl. your 6 new mimeParse cases — nice edge coverage on the body-lookalike and headerless-blob cases).
- ✅ **Live-verified both fixes against real Mail.app**: an Exchange **Sent Items** message that the old unscoped scan could never reach now resolves directly via the `mailbox`+`account` hint, returning the full body *and* `rfcMessageId` (`…@serviceexpress.com`). The `idLocationIndex` design — FIFO-capped, stale-entry falls through to the full scan — is exactly right, and the case-insensitive mailbox iteration (with the documented `mailbox "INBOX" of account` throw) shows real care.

**One thing blocks the merge — the repo's release rule** (CI `require-version-bump` will fail as-is): any PR that changes shipped code must include, in the same PR:
1. A **patch version bump**: `pnpm version patch --no-git-tag-version` (from the repo root), then `node scripts/sync-plugin-version.mjs` to sync the plugin manifests.
2. A **CHANGELOG.md entry** under a new version heading (match the existing style — see `[2.8.6]` for reference): one `### Fixed` bullet for the AppleScript-path Message-ID, one for the scoped Sent-Items fetch.
3. Re-run `pnpm run build` after the bump so the committed bundle carries the new version, and include it in the commit.

Push that to this branch and I'll approve CI + merge same-day. (Rationale for the rule: our publish workflow only releases when the version changes — an un-bumped code change would silently never ship to npm.)

sweetrb:
Merged — thank you, @Goaleve1! 🎉 Two genuinely well-diagnosed fixes: surfacing the RFC Message-ID on the AppleScript path (not just IMAP), and — the one I especially appreciated — the `id → {account, mailbox}` index that lets a by-id fetch open the one right mailbox instead of linear-scanning 700+ mailboxes and timing out before it ever reaches a large Sent Items. The FIFO cap with stale-entry-falls-through-to-full-scan is exactly the right safety valve, and your six `extractRfcMessageIdFromSource` edge cases (body look-alike, headerless blob) were a nice touch.

Since it had been a few days, I went ahead and added the release bits your fork was missing (the mandatory patch bump + CHANGELOG that our `require-version-bump` gate needs, plus a rebuild of the committed bundle against current `main`) and merged it under your authorship. Shipping as **v2.8.9**. Thanks again for a clean first contribution — hope to see more.

PR #97, opened 2026-07-18
fix: enforce mail operation boundaries

jjoanna2-debug:
## Summary

- Bind every composite IMAP message ID to its resolved account identity for reads, attachments, threading, and single/batch mutations. Configured account labels and login addresses remain equivalent selectors; genuinely different accounts fail before any IMAP connection.
- Restrict SMTP `From` overrides to the authenticated user, configured default sender, or aliases explicitly listed in `APPLE_MAIL_MCP_SMTP_ALLOWED_FROM`, with the contract documented in the README, CLI help, and changelog.
- Cap inline Base64 attachments at 25 MiB decoded while accepting standards-wrapped input, clean partial temporary materialization on every exception, strip control characters from Message-ID/account AppleScript literals, and resolve attachment save directories through `realpath` while rejecting existing symlink destinations.
- Rebase onto v2.8.9, bump the package and synchronized plugin manifests to v2.8.10, add complete `[Unreleased]` notes, and rebuild both committed bundles.

These changes harden existing local trust boundaries. They do not claim unauthenticated remote code execution; the relevant cases require a configured mail account and a caller able to invoke the MCP tools.

## Verification

- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run typecheck`
- `pnpm test` — 28 files, 392 tests passed
- `pnpm run test:coverage` — 28 files, 392 tests passed
- `pnpm run test:integration` — 47 tests passed, 10 skipped against configured Mail.app
- `RUN_IMAP_IT=1 GREENMAIL_HOST=127.0.0.1 GREENMAIL_IMAP_PORT=3143 GREENMAIL_USER=tester GREENMAIL_PASS=secret pnpm run test:imap` — checksum-verified GreenMail 2.1.0, 10/10 tests passed
- Focused review suite — 6 files, 120/120 tests passed
- `pnpm run build`, followed by `git diff --exit-code -- build/`
- Second deterministic build produced identical `build/index.js` and `build/cli.js` hashes
- `pnpm pack --dry-run`
- `pnpm audit --audit-level=high` — no known vulnerabilities
- Local version guard — 2.8.9 -> 2.8.10
- `git diff --check`
- Push hook — typecheck and 392 tests passed

## Risk / Notes

- Public-facing and release files changed: `README.md`, `CHANGELOG.md`, `docs/IMAP-SETUP.md`, package/plugin version manifests, and committed `build/` bundles.
- Existing SMTP authenticated-user and configured-default `From` values remain allowed. Calls that relied on arbitrary per-message sender overrides must explicitly configure the intended aliases.
- Attachment save containment rejects stable symlink destinations and canonicalizes the parent directory. As with pathname-based filesystem APIs generally, callers should use directories they control rather than attacker-writable shared directories.
- The fork's `main` branch was synchronized to upstream v2.8.9 (`3d91310`) outside this PR commit. No GitHub repository settings were changed.
- No issue is linked because the current open issue list has no matching report.

## Checklist

- [x] Bug fix
- [x] Tests pass locally
- [x] Linting passes
- [x] Build succeeds
- [x] I have read the contributing guidelines
- [x] The code follows the project's style
- [x] Documentation is updated

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.8.9",
+      "version": "2.8.10",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.8.9",
+      "version": "2.8.10",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.8.9",
+  "version": "2.8.10",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.8.9",
+  "version": "2.8.10",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.8.9",
+      "version": "2.8.10",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.8.9",
+  "version": "2.8.10",
   "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
   "author": {
     "name": "Rob Sweet",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **Attachment reads and saves now enforce filesystem and size boundaries.** Saved attachments are resolved beneath the requested destination directory, reject traversal and existing symlink targets, and route temporary Base64 extraction through a directory rather than a caller-influenced file path. Inline attachments are capped at 25 MiB decoded, partial temporary materialization is cleaned up on failure, and standards-wrapped Base64 is measured without counting decoder-ignored whitespace.
+- **SMTP From overrides are restricted to configured sender identities.** `send-email` email-form `account` values and `apple-mail-send --from` must match the SMTP login user, `APPLE_MAIL_MCP_SMTP_FROM`, or an address in `APPLE_MAIL_MCP_SMTP_ALLOWED_FROM`; rejected identities fail before a transport connection is created.
+
+### Fixed
+
+- **IMAP composite ids cannot be silently routed through a different account.** Single-message, thread, and batch operations now reject a genuinely mismatched account selector while treating a configured account label and its login email as equivalent aliases.
+- **Message-ID lookup uses the shared AppleScript string escaper.** RFC Message-ID values are escaped consistently before interpolation into Mail.app queries.
+
 ## [2.8.9] - 2026-07-18
 ### Fixed
 - **`get-message` now returns the stable RFC Message-ID on the AppleScript path, not just over IMAP.** A prior change surfaced `rfcMessageId` only in the IMAP branch; the Mail.app/AppleScript path still returned `null`, leaving callers with no stable `<…@…>` identifier for dedup/threading. It now reads Mail.app's native `message id` property (normalized, angle brackets stripped) so both backends return the Message-ID. Thanks to @Goaleve1 (#93).
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ Send a new email immediately.
 | `body` | string | Yes | Email body (plain text) |
 | `cc` | string[] | No | CC recipients |
 | `bcc` | string[] | No | BCC recipients |
-| `account` | string | No | Send from specific account (with `transport: "smtp"`, overrides the From address) |
+| `account` | string | No | Mail.app account label, or an email-form SMTP From override. An SMTP override must match `APPLE_MAIL_MCP_SMTP_USER`, `APPLE_MAIL_MCP_SMTP_FROM`, or an address in `APPLE_MAIL_MCP_SMTP_ALLOWED_FROM` |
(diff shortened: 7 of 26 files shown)


sweetrb (changes requested):
Thanks for this, @jjoanna2-debug — it's a substantial, well-structured hardening pass, and several pieces are exactly right: the attachment-save containment (realpath + out-of-root + existing-symlink rejection, and routing `getAttachmentBase64` through the temp *directory* rather than a file path) is sound and well-covered by the new tests; the 25 MiB decoded cap with partial-materialization cleanup is a good add; and switching `findNumericIdByMessageId`'s quoting to `escapeForAppleScript` is the right call. I ran a full multi-pass review (correctness / regression / security-tests-docs) with adversarial verification of each finding against the live diff and a real multi-account config.

**Verdict: needs changes — and it can't merge as-is** (currently `BLOCKED`). What's required:

**1. (Blocking) Version bump + CHANGELOG + bundle rebuild.** The diff changes six non-test `src/**` files but not `package.json` or `CHANGELOG.md`, so the required `require-version-bump` check fails. `main` has also since advanced to **2.8.9**, so this branch is now behind — please **rebase onto current `main`**, bump to **2.8.10** (`pnpm version patch --no-git-tag-version` + `node scripts/sync-plugin-version.mjs`), add a `CHANGELOG` `[Unreleased]` entry describing each behavior change, and **rebuild the committed bundle** (`build/index.js` + `build/cli.js`) in the same commit — main's 2.8.9 touched `appleMailManager.ts`, so the bundle will otherwise fail the "Verify committed build/ matches source" step.

**2. (Should fix) The IMAP account-binding guard rejects a valid account selector.** `depsForAccount` compares the caller's `account` to the id's encoded account with raw string equality, but ids are encoded with the account *label* (`cfg.accountLabel`, `imapClient.ts:499-500`) while `resolveImapConfig`/`isImapAccount` intentionally accept **either the label or the login `user`** as a valid selector (`imapClient.ts:276,340`). So on a config where label ≠ login-address, `get-thread` and `batch-move-messages` called with the *email-form* account against a label-encoded id now throw `IMAP message id belongs to account "<label>", not "<email>"` — a previously-working call. (The genuinely-different-account case you're guarding is an improvement — the old `deps.account ?? ref.account` silently ran against the wrong account, whereas now it errors loudly. It's specifically the label-vs-login-email equivalence that regresses.) Please compare *resolved* account identity (accept either alias) instead of raw strings, and add a test for the email-form selector — the new tests only exercise the genuine-mismatch deny path.

**3. (Should fix) Docs/behavior drift on the new SMTP `From` allowlist.** `sendViaSmtp` now hard-fails any `From` not in `{user, configured-from, APPLE_MAIL_MCP_SMTP_ALLOWED_FROM}` before creating the transport. That's a behavior change to two documented send surfaces — `send-email` with `account=<alias>` (mapped to `From` when it contains `@`) and the `apple-mail-send` CLI, which *requires* `--from` and passes it verbatim — and the docs still describe the old pass-through behavior: the README `account` param row + SMTP-transport bullet, and the `src/cli.ts` `--from` USAGE line (not touched in this PR — only the compiled `build/cli.js` was rebuilt, so `--help` now misstates the contract). We're evaluating the disposition of this tightening; if it stays, please document it (including `APPLE_MAIL_MCP_SMTP_ALLOWED_FROM`) across the README, `src/cli.ts` USAGE, and the CHANGELOG.

**4. (Minor) Base64 size pre-check false-rejects standards-wrapped input.** `MAX_INLINE_ATTACHMENT_BASE64_CHARS = ceil(bytes/3)*4` is the *unwrapped* ceiling, but the char pre-check (and the mirrored zod `.max`) count whitespace while `Buffer.from` ignores it when decoding — so RFC-2045 line-wrapped base64 for a decoded size in the ~24.4–25 MiB band is rejected before the authoritative decoded-byte check ever runs. Narrow (agents almost always emit unwrapped base64), but worth stripping whitespace before measuring, or relaxing the char ceiling.

The security intent is good and most of the mechanics are solid — happy to help get it over the line. Thanks again.

jjoanna2-debug:
Thanks for the detailed review. I addressed all four items in `a188a15665c23fce4212879da095b684b952986a`:

1. Rebasing/version/build: rebased onto current `main` (`3d91310`, v2.8.9), bumped the package and plugin manifests to 2.8.10, added the `[Unreleased]` changelog entries, and rebuilt both committed bundles in the same commit.
2. IMAP account identity: account binding now treats a configured label and its login address as aliases of the same resolved account, while still rejecting genuinely different accounts. Added regression coverage for both single-message retrieval and batch move using the email-form selector against label-encoded IDs; the mismatch tests remain in place.
3. SMTP `From` contract: documented the authenticated user/configured sender/`APPLE_MAIL_MCP_SMTP_ALLOWED_FROM` allowlist and reject-before-connect behavior in the README, source CLI usage, and changelog. Added a CLI help assertion for the environment variable.
4. Wrapped Base64: the cheap precheck is now whitespace-aware and separately raw-bounded, while decoded bytes remain the authoritative 25 MiB limit. Added a near-limit RFC 2045 CRLF-wrapped regression test whose raw length exceeds the old ceiling.

Verification on the rebased head:

- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run typecheck`
- `pnpm test` — 392/392
- `pnpm run test:coverage` — 392/392
- `pnpm run test:integration` — 47 passed, 10 skipped
- `RUN_IMAP_IT=1 ... pnpm run test:imap` against checksum-verified GreenMail 2.1.0 — 10/10
- `pnpm run build`, followed by a second build with identical `build/index.js` and `build/cli.js` hashes
- `pnpm pack --dry-run`
- `pnpm audit --audit-level=high` — no known vulnerabilities
- version guard: 2.8.9 -> 2.8.10
- `git diff --check`
- `git diff --exit-code -- build/`

The branch is force-pushed with lease and is mergeable; it is ready for re-review.

sweetrb (approved):
Re-reviewed at `a188a15` — all four items from the prior pass are resolved, and I verified each against the source rather than the summary:

1. **Version + CHANGELOG + bundle** — 2.8.9 → 2.8.10 across `package.json` and every plugin/marketplace manifest in lockstep; `[Unreleased]` Security/Fixed entries are specific and accurate. Rebuilt the committed bundle from a clean checkout: `git diff --exit-code -- build/` is clean, so `build/index.js` + `build/cli.js` reproduce exactly from source.
2. **IMAP account identity** — `sameImapAccount` now resolves both selectors against the config and treats `accountLabel` and login `user` as aliases of the same account, so the email-form selector against a label-encoded id is accepted while a genuinely different account still throws loudly. Regression coverage for the email-form path (single + batch) is present alongside the mismatch-deny tests.
3. **SMTP From allowlist** — documented across the README, `src/cli.ts` USAGE, and the CHANGELOG; the zod schema uses the relaxed input bound plus the shared `isInlineAttachmentBase64WithinLimit` refine, so the tool contract and `--help` now match the runtime.
4. **Wrapped base64** — the pre-check is whitespace-aware at both the util layer and the zod layer (raw `INPUT_CHARS` bound + non-whitespace count against the encoded ceiling), with the decoded-byte limit still authoritative. New near-limit RFC-2045 CRLF-wrapped test confirms it.

Local verification on the PR head: lint / format:check / typecheck clean, `pnpm test` 392/392, `pnpm run test:integration` 47 passed / 10 skipped against real Mail.app, bundle reproduces from source. The changes are strictly defensive — no new dependencies, no network calls, no widened data flow.

The SMTP From-restriction disposition is confirmed: it stays. Approving. Thanks for the thorough follow-up — the account-alias fix in particular was exactly the right resolution.

sweetrb:
Merged as v2.8.10 — thanks, @jjoanna2-debug. This was a genuinely strong hardening pass, and your follow-up nailed all four review items:

- The **account-identity fix** (`sameImapAccount` resolving `accountLabel` and login address as aliases of the same resolved account) was exactly the right resolution — it preserves the loud rejection of a genuinely different account while unbreaking the email-form selector against a label-encoded id, with regression coverage for both the single and batch paths.
- The **whitespace-aware base64 pre-check** at both the util layer and the zod `refine`, keeping the decoded-byte limit authoritative, is clean — and the near-limit RFC-2045 CRLF-wrapped test is a nice touch.
- The **attachment-save containment** (realpath canonicalization + out-of-root + existing-symlink rejection, and routing `getAttachmentBase64` through the temp *directory*) and the **SMTP From allowlist** with its documented `APPLE_MAIL_MCP_SMTP_ALLOWED_FROM` escape hatch are both well done.

I verified locally on your head: 392 unit + 47 integration tests green, and the committed bundle reproduces exactly from source. The one CodeQL "insecure temporary file" flag is a false positive — `mkdtempSync` creates an owner-only (0700) randomly-named directory, and your change additionally hardened that save path. Appreciate the thorough verification write-up too.

PR #105, opened 2026-07-21
feat: add smart mailbox tools (list/create/delete + newsletter auto-detect)

maf4711:
Smart mailboxes are criteria-based virtual views in Apple Mail, so they
can't be scripted via AppleScript's normal mailbox terms (unreliable on
German-localized macOS). Manipulates the SyncedSmartMailboxes.plist
directly instead. Adds create-newsletter-smart-mailboxes to scan the
Inbox for newsletter senders (List-Unsubscribe, noreply, repetitive
subjects) and auto-create dedicated smart views for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -879,6 +879,59 @@ Rename a mailbox (creates new, moves messages, deletes old).
 
 ---
 
+### Smart Mailbox Operations (intelligente Postfächer)
+
+Smart mailboxes are virtual (criteria-based views, not real folders). These tools let you create dedicated smart views for newsletters etc. directly from the Inbox without moving messages.
+
+#### `list-smart-mailboxes`
+
+List existing smart mailboxes.
+
+**Parameters:** None
+
+**Returns:** List of smart mailbox names + criteria summary.
+
+---
+
+#### `create-smart-mailbox`
+
+Create a smart mailbox with a simple contains rule.
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `name` | string | Yes | Name for the smart mailbox |
+| `fromContains` | string | No | Match if From contains this |
+| `subjectContains` | string | No | Match if Subject contains this |
+| `bodyContains` | string | No | Match if Body contains this |
+
+Exactly one of the three contains fields should be provided.
+
+---
+
+#### `delete-smart-mailbox`
+
+Delete a smart mailbox by name.
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `name` | string | Yes | Smart mailbox name |
+
+---
+
+#### `create-newsletter-smart-mailboxes`
+
+High-level tool: scan recent messages in your INBOXes, detect likely newsletters (volume + signals like List-Unsubscribe, noreply, repetitive subjects), and create smart mailboxes for them (names prefixed "NL: ...").
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `dryRun` | boolean | No | Default true — only propose, do not create |
+| `minCount` | number | No | Min messages from a sender (default 3) |
+| `days` | number | No | Lookback window in days (default 90) |
+
+Use with `dryRun: false` to actually create the smart folders for newsletters cluttering your Inbox.
+
+---
+
 ### Account Operations
 
 #### `list-accounts`
--- a/build/index.js
+++ b/build/index.js
@@ -75689,13 +75689,16 @@ import {
   existsSync as existsSync3,
   writeFileSync as writeFileSync3,
   readFileSync as readFileSync2,
+  readdirSync as readdirSync2,
+  unlinkSync,
   mkdtempSync as mkdtempSync2,
   rmSync as rmSync2,
   realpathSync,
   lstatSync
 } from "fs";
 import { isAbsolute, resolve, sep, join as join4 } from "path";
-import { homedir as homedir3 } from "os";
+import { homedir as homedir3, tmpdir as tmpdir2 } from "os";
+import { randomUUID } from "crypto";
 
 // src/utils/applescript.ts
 import { execSync, spawnSync } from "child_process";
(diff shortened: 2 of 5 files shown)


sweetrb (changes requested):
Thanks for this, @maf4711 — the German-localization diagnosis is spot on. AppleScript's `smart mailbox` / `intelligentes Postfach` terms genuinely don't compile reliably on localized macOS, so going straight to `SyncedSmartMailboxes.plist` is a legitimate way around it, and the reverse-engineered criteria structure is careful work. I also like that `create-newsletter-smart-mailboxes` defaults to `dryRun: true` and that the scoring blends volume with real signals (List-Unsubscribe, `noreply`, repetitive subjects).

I can't merge it as-is, though — there are a few blocking issues, mostly around safety and the repo's release gates. Verdict: **needs changes**.

### Blocking

1. **`killall Mail` is a silent, destructive side effect.** `createSmartMailbox` kills the user's Mail.app with no warning or opt-in. That can drop an unsaved draft, and it interrupts *this MCP server's own* AppleScript session. Worse, `killall` sends `SIGTERM`, so Mail quits **gracefully** and flushes its in-memory mailbox state to disk — which can overwrite the plist edit you just wrote (a classic lost-update race). Editing `SyncedSmartMailboxes.plist` while Mail is running is inherently racy. Please drop the `killall` entirely; if a reload is needed, either only write when Mail isn't running, or just document that the new smart mailbox appears the next time Mail is launched. Any process-killing must be opt-in and disclosed.

2. **The whole-plist `plutil` JSON round-trip is lossy and can corrupt *existing* smart mailboxes.** `loadSmartMailboxes`/`saveSmartMailboxes` convert the entire file `binary1 → json → (edit) → binary1`. JSON has no `<data>` or `<date>` type, so any *other* smart mailbox in the file whose criteria use those (e.g. date-based rules) gets silently rewritten as a plain string on save — you'd be corrupting the user's other smart mailboxes, not just adding one. On an English library the file often happens to be all strings/bools/ints and round-trips clean, but that isn't guaranteed. Please use a lossless path: round-trip via `plutil -convert xml1` (xml1 preserves `<data>`/`<date>`), or append the single entry with `PlistBuddy` instead of rewriting the whole file.

3. **No tests.** This is 777 additions with zero coverage, and CI gates on `test (22)` / `test (24)`. The pure pieces are easy to cover if you factor the AppleScript exec out of `findNewsletterCandidates`: `buildSmartMailboxEntry` (criteria shape + `From`/`Subject`/`Body` header selection), `extractEmail`, and the scoring/grouping. Please add unit tests for those.

4. **Tool registration doesn't match the repo convention and ships no descriptions.** Every other tool here uses `server.registerTool(name, { description, inputSchema, outputSchema }, handler)` with a structured `Use when: / Returns: / Do not use when:` description — and a `Safety:` line for write/destructive tools (see `delete-note`). These four use `server.tool(name, shape, handler)`, a different API with **no description at all**, so an agent sees them with zero guidance. `create-smart-mailbox`, `delete-smart-mailbox`, and especially `create-newsletter-smart-mailboxes` (writes files) need the `Safety:` callout. Please convert to `registerTool` + structured descriptions.

5. **Missing version bump + CHANGELOG (release gate).** This changes shipped bytes (`src/**` and the committed `build/index.js`), so `version-guard` will fail until `package.json` gets at least a patch bump and `CHANGELOG.md` gets an `[Unreleased]` → `### Added` entry in the same PR. Also: the committed `build/index.js` in this PR was built in your fork, and fork PRs don't get our automatic bundle rebuild — CI's "Verify committed build/ matches source" step re-runs `pnpm run build` and diffs, so the bundle must be rebuilt from a clean `pnpm install --frozen-lockfile` (not a symlinked `node_modules`) or it'll fail even with identical source.

### Non-blocking notes

- `findNewsletterCandidates` reads `source of m` (the full RFC822 source) for up to 400 messages per inbox — that's the exact per-message Apple Event storm that's caused timeouts elsewhere in this project (cf. the sibling notes repo's #100). Even with the 120s timeout it may not finish on a large inbox. Consider bulk-fetching `sender`/`subject` and only sampling `source` for the top-N grouped candidates.
- `createSmartMailbox` returns `true` both when it creates the mailbox and when it already existed; the "created (or already existed)" message hides the distinction, so a caller can't tell whether anything changed.
- The temp filenames use `Date.now()`; two calls in the same millisecond collide. `randomUUID()` (already imported) would be safer.

Once the `killall` / plist-safety / tests / registration items are in, I'm happy to help with the mechanical release bits (version bump, CHANGELOG, bundle rebuild) — "Allow edits from maintainers" is on, so I can push those to the branch for you at that point. Thanks again for tackling a genuinely awkward corner of Mail scripting.

sweetrb (changes requested):
Thanks for this, @maf4711 — the core idea is genuinely good. You're right that AppleScript's `smart mailbox` terms are unreliable on localized macOS, and going through `SyncedSmartMailboxes.plist` is a sound workaround. The newsletter auto-detection is a nice touch. I'd like to land it, but a few things need to change first — mainly around **not risking the user's existing smart mailboxes.**

### Blocking — data-loss risk in `saveSmartMailboxes`

```js
const r = spawnSync("plutil", ["-convert", "binary1", "-o", plistPath, tmpJson], …);
```

This rewrites the **entire** live `SyncedSmartMailboxes.plist` in place, and it's the whole plist — every existing smart mailbox, not just the new one. Three concrete hazards:

1. **No backup.** If anything goes wrong, the user's existing smart mailboxes are gone with no recovery. Please copy the original to `SyncedSmartMailboxes.plist.bak` (or a timestamped backup) before the first write.
2. **Non-atomic.** `plutil -o plistPath` writes directly to the real file; an interrupted write truncates it. Write to a temp plist and `fs.renameSync()` into place (atomic on the same filesystem) so a failure leaves the original intact.
3. **Lossy round-trip.** `loadSmartMailboxes` does `plutil -convert json` → you edit → `plutil -convert binary1`. Plist `<data>` and `<date>` values don't survive a JSON round-trip cleanly, and smart-mailbox *criteria* can contain dates (e.g. "date received before X"). Because you rewrite the whole plist, a lossy round-trip could corrupt criteria in **other** mailboxes the user never asked you to touch. Worth either preserving the raw plist for untouched entries, or verifying criteria survive the round-trip on a plist that has date-based rules.

### Blocking — `killall Mail`

`spawnSync("killall", ["Mail"])` force-quits Mail with no warning, which loses any open compose windows. Please make this opt-in (a parameter, default off) or at least document it loudly. A gentler reload path is preferable if one exists.

### Required before merge

- **Version bump + CHANGELOG.** This is a feature, so it needs a `pnpm version minor` + a `## [Unreleased]`/dated CHANGELOG entry in the same PR — `version-guard` will hard-block otherwise.
- **CI isn't running.** The PR head is your fork's `main` branch, so the required checks never triggered. Pushing the work to a topic branch (e.g. `feat/smart-mailboxes`) and repointing the PR will let CI run.
- **`delete-smart-mailbox` needs a `Safety:` callout** in its MCP tool description (it deletes a user's mailbox), per this project's tool-description convention (`Use when: / Returns: / Do not use when: / Safety:`).
- **Real-data testing.** Because this writes the actual Mail plist, I'll want to verify a create/delete round-trip against a real library with pre-existing smart mailboxes (including a date-criteria one) before merging — happy to do that once the backup/atomic changes are in.

If `maintainerCanModify` stays on I'm glad to help with the mechanical parts (version bump, the atomic-write/backup helper) once you've had a look at the round-trip fidelity question, which is the one that really needs your eyes. Really nice contribution — just want it to be safe to run against a populated Mail.

sweetrb:
Thanks so much for this, @maf4711 — your German-localization diagnosis was the key insight, and going straight to `SyncedSmartMailboxes.plist` (rather than fighting AppleScript's `smart mailbox` / `intelligentes Postfach` terms) is exactly the right workaround. The newsletter auto-detection with its dry-run default and multi-signal scoring is a genuinely nice touch.

This has now **shipped in v2.9.0** 🎉 — your commit was carried forward in #110 with your authorship preserved via `Co-authored-by`. I opened a continuation PR rather than merging here only for a mechanical reason: the PR head was your fork's `main` branch, so our required CI checks never triggered on it and there was no way to repoint it.

On top of your feature, #110 added the safety/release pieces we talked about:
- **Lossless, non-destructive plist writes** — it turned out the whole-file `plutil -convert json` round-trip doesn't just lose `<data>`/`<date>`, it *fails outright* on any plist that contains them, which would have wiped a user's existing smart mailboxes on a library with a date-based rule. Writes now append/delete a single entry (`plutil -insert -json` / `PlistBuddy Delete`) on a backed-up temp copy, `plutil -lint` it, then atomically rename into place.
- Dropped `killall Mail` (the lost-update race), converted the four tools to `registerTool` with structured descriptions + `Safety:` callouts, and added tests including real-plist round-trips that prove date/data entries survive.

I verified the whole thing against a real Mail library (create → 15 mailboxes, all originals intact → delete → back to exactly the original set, byte-for-byte) before release. Really appreciate the contribution — thank you!

PR #115, opened 2026-07-29
chore(hermes): remove hermes-plugin packaging docs

maf4711:
## Summary

Removes the `.hermes-plugin/` packaging docs (`README.md`, `config.yaml`) from this repo.

Hermes Agent has no plugin/marketplace drop-in; the previous packaging docs were easy to misread as installable manifests. Users should configure Hermes via `hermes mcp add` (or merge config into `~/.hermes/config.yaml`) instead of treating `.hermes-plugin/` as a drop-in.

## Test plan

- [x] Repo builds / tests green locally (`tsc --noEmit`, vitest)
- [ ] Confirm README “Other Hosts” / Hermes docs still point at the right setup path after this lands (follow-up if links still reference deleted files)


sweetrb (changes requested):
Thanks for this, @maf4711 — and for flagging the README follow-up yourself in the test plan; that instinct was right, and it turns out to be one of three things that need to land in the same change.

Holding off on merge for now. Details below.

### Blocking: this would break the release flow

`package.json`'s `version` lifecycle script stages the plugin manifest dirs by name:

```json
"version": "node scripts/sync-plugin-version.mjs && git add .claude-plugin .agents/plugins codex .hermes-plugin .antigravity-plugin",
```

`git add` on a pathspec that matches nothing is a fatal error, not a no-op:

```
$ git add a.txt .hermes-plugin
fatal: pathspec '.hermes-plugin' did not match any files   # exit 128
```

So once `.hermes-plugin/` is gone, `pnpm version <patch|minor|major> --no-git-tag-version` — the documented first step of every release here — fails, and nothing can ship until someone edits `package.json`. Removing `.hermes-plugin` from that `git add` list has to be part of this PR.

Worth noting neither of us would have caught this from the checks: `tsc --noEmit` and vitest don't exercise the release path, and CI doesn't either.

(`.github/workflows/dependabot-rebuild.yml:243` names the dir too, but it's `2>/dev/null || true`-guarded, so that one degrades quietly rather than breaking.)

### Blocking: two dead README links

`README.md:83` links directly at both deleted files:

> …or merge [`.hermes-plugin/config.yaml`](…) into `~/.hermes/config.yaml`. Details: [`.hermes-plugin/README.md`](…).

That bullet needs rewriting so the `hermes mcp add` command and the `mcp_servers:` snippet live inline in the README.

### Direction question — worth settling before the work

Two things make me want your read before you spend more time here:

1. **All four Apple MCP servers carry `.hermes-plugin/`** (mail, notes, numbers, photos). Removing it from mail alone breaks a deliberate multi-host packaging parity across the family, so this is really a four-repo change or a none-repo change.
2. **The "easy to misread as installable manifests" problem was already fixed once**, deliberately. The current `.hermes-plugin/README.md` opens with "Hermes doesn't install plugins from a repository — MCP servers are registered in `~/.hermes/config.yaml`, or via the `hermes mcp` CLI. There is no `plugin.json` / `marketplace.json` to point at", and `config.yaml`'s header comment repeats it. An earlier release replaced Claude-format JSON with exactly this wording for that reason.

So the current files may already be doing the job the deletion is aiming at. If you're still hitting real-world confusion despite that wording, I'd genuinely like to hear the specifics — that's a better bug than the packaging itself, and it might point at a README fix rather than a deletion.

### Mechanical note, not your fault

This PR's head branch is your fork's own `main`, which means required CI never triggers on it — `gh pr checks` reports "no checks reported", and a PR's head can't be repointed after creation. Even with everything above fixed it can't merge in this form; I'd carry the commits onto an in-repo topic branch as a continuation PR (crediting you via `Co-authored-by`) so full CI runs. Easiest fix on your side for future PRs is a topic branch on the fork rather than `main`.

### Smaller

- Needs a CHANGELOG entry under `## [Unreleased]` (`### Removed` or `### Documentation`). No version bump needed — `.hermes-plugin` isn't in `package.json` `files[]`, so nothing shipped to npm changes.

Happy to take the whole thing off your hands and land it across all four repos with attribution if you'd rather — just say the word. Either way I want to settle the direction question first.

sweetrb:
Thanks @maf4711 — you were right that `.hermes-plugin/` read like an installable manifest when Hermes has no drop-in to install from. That directory was a leftover from when it *did* ship Claude-format JSON, and trimming it is the right call.

I pushed one follow-up commit to your branch (`cb6150c`) rather than sending this back, because the removal had two loose ends — one of which you already flagged yourself:

**1. `pnpm version` would have broken on every future release.** `package.json`'s `version` lifecycle script still listed the deleted directory:

```
"version": "node scripts/sync-plugin-version.mjs && git add .claude-plugin .agents/plugins codex .hermes-plugin .antigravity-plugin"
```

`git add` exits **128** on a pathspec that matches nothing, so `pnpm version <patch|minor|major> --no-git-tag-version` — the documented release step for this repo — would have failed as soon as this landed. A nice quiet landmine; verified before and after (old list → 128, new list → 0). Dropped the stale path.

**2. The README links you flagged.** Your test plan had "confirm README Other Hosts / Hermes docs still point at the right setup path" unchecked, and it was a real hit: the Hermes bullet linked to both deleted files, so they would have 404'd post-merge. Rather than just deleting the links, I moved the useful content out of the doomed README and inline into the main one — the `hermes mcp add` command, the `~/.hermes/config.yaml` `mcp_servers:` snippet, and the restart note. Arguably better placed there anyway, since nobody browses a dot-directory for setup instructions.

Plus CHANGELOG entries under `[Unreleased]`.

**Deliberately left alone:** `.github/workflows/dependabot-rebuild.yml` also lists `.hermes-plugin` in a `git add`, but it is guarded by `2>/dev/null || true` so it cannot fail — and editing a workflow file inside a fork PR makes it unmergeable with our token, besides breaking the four-repo workflow conformance check. That goes in as a separate in-repo change.

No version bump owed: `.hermes-plugin/` was never in `package.json` `files[]`, there is no `src/` change, and the committed bundle is byte-identical (verified with a rebuild). Local gates all green — typecheck, 408 tests, prettier.

Verdict: **merge**.

sweetrb:
Landed as #116 — merged to `main` in 49f1395 with a `Co-authored-by: Marco Föllmer <marco@merados.com>` trailer, so this is credited to you on the commit.

Why a second PR rather than merging this one: your branch was green on all seven required checks, but it is a fork PR whose head is the fork's own default branch, and against this repo's strict (require-up-to-date) branch protection that combination reports `BLOCKED` even with every gate passing. The documented escape is an admin override, and I would rather not bypass protection that is deliberately enforced here — so #116 carries your commit forward cherry-picked (authorship preserved) with content **byte-identical** to the CI-validated head of this PR. Nothing of yours was rewritten or dropped.

Thanks for this one specifically — `.hermes-plugin/` was a leftover from when it genuinely did ship Claude-format JSON manifests, and you were right that leaving a directory of manifest-looking files for a host with no drop-in is worse than having nothing there. Two things fell out of it that were worth more than the cleanup itself:

- `package.json`'s `version` lifecycle script still ran `git add … .hermes-plugin …`. `git add` exits **128** on a pathspec that matches nothing, so `pnpm version` — the documented release step for this repo — would have failed on the very next release. That would have surfaced as a mystifying release breakage days later, disconnected from this change.
- Your own test plan flagged the README as an open item, and it was a real hit: the "Other Hosts" bullet linked to both files you deleted. Instead of just dropping the links I moved the content out of the doomed README and inline into the main one — the `hermes mcp add` command, the `~/.hermes/config.yaml` `mcp_servers:` snippet, and the restart note. Better placed there anyway.

Noting for completeness: the other three Apple MCP servers (notes, numbers, photos) still carry `.hermes-plugin/`, so mail is now the odd one out. Same argument applies to them and I'll raise it separately rather than fold a three-repo packaging change into this.

Second good catch from you this month, after the smart-mailbox work in #105/#110. Appreciated.

PR #123, opened 2026-07-31
Complete brace-expansion sequence bounds

### LINKED ISSUE #121: brace-expansion: lift to 1.1.18 / 5.0.9 once they clear the 24h minimumReleaseAge gate
#119 moved the dev-only `brace-expansion@1` path (ESLint → `minimatch@3.1.5`) onto `1.1.17`, the upstream v1 backport for **GHSA-mh99-v99m-4gvg / CVE-2026-14257**. That fix is **incomplete on both lines**, and the completions landed the same morning #119 was reviewed.

## What 1.1.17 / 5.0.8 missed

`1.1.17` added `EXPANSION_MAX_LENGTH` to bound the accumulator in `combine`, but did **not** thread `maxLength` into `expandSequence` — so the sequence path (`{1..N}`, `{a..z..k}`) stayed uncapped in total output length, which is the same unbounded-memory shape the CVE describes.

`1.1.18` (published 2026-07-30T10:17Z) fixes it:

```js
-function expandSequence(body, isAlphaSequence, max)
+function expandSequence(body, isAlphaSequence, max, maxLength)
...
+    if (length + c.length > maxLength) break
     N.push(c)
+    length += c.length
```

plus bounding the `values` accumulation and correcting `max` accounting for dropped empties.

`5.0.9` (published 2026-07-30T10:00Z) is the **same change** on the v5 line — verified by diffing the two tarballs. So this repo's *other* path, `minimatch@10.2.6 → brace-expansion@5.0.8`, is also on the incomplete fix today.

## Why this was not just folded into #119

`minimumReleaseAge: 1440` (24h). Both releases were ~1 hour old at review time, and `brace-expansion` is a high-value supply-chain target — adding a `minimumReleaseAgeExclude` carve-out to pull a one-hour-old release is exactly the case the gate exists to prevent. Waiting costs nothing real here: the advisory is **dev-toolchain-only** (`brace-expansion` does not appear in `build/index.js`), and reaching it requires feeding adversarial brace input to a lint run.

## To do (from 2026-07-31 ~10:17Z, when both clear the gate)

- [ ] `pnpm-workspace.yaml`: `brace-expansion@<1.1.18: 1.1.18` (replacing the `<1.1.17: 1.1.17` override)
- [ ] Lift the v5 path to `5.0.9` — confirm whether `minimatch@10.2.6`'s range picks it up on a plain re-resolve, or whether it needs its own floor
- [ ] Re-resolve the lockfile with **no** `minimumReleaseAgeExclude` entry, and confirm `pnpm install --lockfile-only` reports "Lockfile passes supply-chain policies"
- [ ] Confirm `build/` stays byte-identical (dev-only ⇒ no version bump), and CHANGELOG the supersession
- [ ] Same pass in the three sibling repos if they resolve `brace-expansion` on either line

Note GitHub's advisory metadata still lists `5.0.8` as the patched version (last updated 2026-07-24), so `pnpm audit` output is not a reliable signal here — compare the published tarballs.

Follow-up to #119 (thanks @jjoanna2-debug).

jjoanna2-debug:
## Summary

- Completes the dev-only brace-expansion remediation by lifting the CommonJS-compatible v1 floor from `1.1.17` to `1.1.18` and the independent v5 path from `5.0.8` to `5.0.9`.
- Adopts both releases only after they cleared the repository 1440-minute `minimumReleaseAge` gate, with no `minimumReleaseAgeExclude` and no audit suppression.
- Records the completed sequence-path bound under CHANGELOG `[Unreleased]`; the shipped bundle remains byte-identical.

## Verification

- `pnpm install --frozen-lockfile` — lockfile passes supply-chain policies (369 entries).
- `pnpm test` — 29 files, 408 tests passed.
- `pnpm run lint` — 0 errors; 10 pre-existing warnings.
- `pnpm run typecheck`.
- `pnpm run format:check`.
- `pnpm run build`.
- SHA-256 before/after build: `build/index.js` `5e68e0d25b68475052acd9d1d39a6b8e7c1fb43f0723d8457c5b4fcbf96e12af`; `build/cli.js` `b26026b4ba98c49ee6074b836ef917b5e6d348e6cddc741943db2f902bca8617`.
- `git diff --exit-code -- build/`.
- `git diff --check`.

## Risk / Notes

- Public-facing `CHANGELOG.md` changes under `[Unreleased]`.
- Dev-toolchain dependency only; no runtime source, runtime dependency, package version, or shipped bytes change.
- Closes #121 after merge.


sweetrb:
Merged — thanks @jjoanna2-debug, and thanks especially for holding this to the 24-hour gate rather than reaching for a `minimumReleaseAgeExclude` carve-out. Pulling a one-hour-old `brace-expansion` release behind an exclusion is exactly the scenario the gate exists to prevent, and "wait a day" was the right call for a dev-only advisory that isn't in the shipped bundle.

The 10,000-zero pad in your repro harness was the detail that made this checkable. My own 60-character measurement showed only a ~30% RSS difference and I couldn't reproduce your magnitudes; the missing parameter turned it into 4,606 ms → 9 ms and 176 MB → 61 MB, with byte-identical final output. That's what separated "the diff plainly adds a bound" from "the bound demonstrably matters", and it's why this landed on evidence instead of on the advisory text — which is still wrong in both directions (GitHub lists `5.0.8` as first-patched, and its `<= 5.0.7` range marks the entire v1 line vulnerable under semver).

Two things I pushed to your branch before merging:

**A floor for the v5 line.** The lift moved v5 to `5.0.9`, but only by re-resolution — `brace-expansion@<1.1.18` doesn't match `5.0.8` under semver, so nothing actually pinned that result. Both majors now carry an explicit two-sided floor. They have to be two-sided: a bare `<5.0.9` also matches `1.1.18`, and would force minimatch 3's CommonJS path onto the v5 ESM API — the same `expand is not a function` failure that made the v1 backport necessary in #119.

**One `[Unreleased]` security entry instead of two.** Neither had shipped, so the release notes would have carried "deliberately partial, deferred until the gate clears" directly above "completed" for the same version.

Your finding also turned out to be bigger than this repo. Checking the siblings for #121's last checklist item: all three resolve a **third** major (`minimatch@9` → v2) that mail doesn't have, and two of their four paths (`1.1.16`, `5.0.7`) were below even the advisory's floor. Ported with three floors each — sweetrb/apple-notes-mcp#113, sweetrb/apple-numbers-mcp#49, sweetrb/apple-photos-mcp#60.

PR #148, opened 2026-08-12
fix: advertise tool schemas as JSON Schema 2020-12, not draft-07 (#147)

### LINKED ISSUE #147: Tools fail in Claude Desktop — outputSchema is draft-07, client now requires JSON Schema 2020-12
Since a recent Claude Desktop update, every tool call fails before executing with:

> Tool '<name>' has an invalid outputSchema: JSON Schema declares an unsupported dialect ("$schema": "http://json-schema.org/draft-07/schema#"). The default validator supports JSON Schema 2020-12 only.

It affects all tools (list-messages, get-message, batch-move-messages, batch-flag-messages, batch-delete-messages, create-draft, etc.), so the server is currently unusable in Claude Desktop.

**Cause:** The server emits tool schemas in JSON Schema draft-07 (the zod-to-json-schema default). MCP has since standardized on 2020-12 as the required dialect (SEP-834 / SEP-1613 / SEP-2106), and clients now enforce it.

**Fix:** Update schema generation to emit 2020-12 — e.g. move to the newer MCP TypeScript SDK's Standard Schema / Zod v4 support (z.toJSONSchema({ target: 'draft-2020-12' })), or configure zod-to-json-schema for the 2020-12 target. Tracked upstream in modelcontextprotocol/typescript-sdk#745.

**Env:** Claude Desktop (macOS), installed via `npx -y apple-mail-mcp`.

sweetrb:
Fixes #147.

Every tool is currently refused by JSON Schema 2020-12 clients:

```
Tool '<name>' has an invalid outputSchema: JSON Schema declares an unsupported
dialect ("$schema": "http://json-schema.org/draft-07/schema#"). The default
validator supports JSON Schema 2020-12 only.
```

The server connects; nothing is callable. All 50 tools are affected, on both `inputSchema` and `outputSchema`.

## Root cause

It is not our schemas — it is the dialect they are advertised under.

`@modelcontextprotocol/sdk`'s `server/mcp.js` calls `toJsonSchemaCompat(obj, { strictUnions, pipeStrategy })` with **no `target`**. In `sdk/dist/esm/server/zod-json-schema-compat.js`, `mapMiniTarget(undefined)` returns `'draft-7'`, and the zod v3 branch hands off to `zod-to-json-schema`, which defaults to draft-07 and stamps `"$schema": "http://json-schema.org/draft-07/schema#"` onto every emitted schema. MCP has since standardized on JSON Schema 2020-12 (SEP-834 / SEP-1613 / SEP-2106), so a conforming client rejects each tool outright and the whole tool list disappears.

## A zod v4 upgrade does NOT fix this

Verified empirically against **SDK 1.30.0 + zod 4.4.3**: the emitted `$schema` is still draft-07. The SDK never passes a target, so *both* converter branches — v3 (`zod-to-json-schema`) and v4 (`zod/v4-mini` `toJSONSchema`) — fall back to `draft-7`. Upgrading zod changes which converter runs, not which dialect it targets. Please don't re-chase that route.

## The fix

New `src/utils/jsonSchemaDialect.ts` normalizes the outgoing `tools/list` payload at the **transport boundary** — the only public seam that does not reach into SDK internals — and `src/index.ts` wraps the transport:

```ts
const transport = withJsonSchema2020_12(new StdioServerTransport());
```

No tool definition, zod schema, or handler was touched.

It is a real dialect conversion, not a string swap on `$schema`:

| draft-07 | 2020-12 |
|---|---|
| `definitions` | `$defs` (and `#/definitions/X` refs repointed to `#/$defs/X`) |
| tuple-form `items: [A, B]` | `prefixItems` |
| `additionalItems` (alongside a tuple) | `items` |
| `dependencies` | `dependentRequired` / `dependentSchemas` |
| boolean `exclusiveMinimum` / `exclusiveMaximum` + `minimum` / `maximum` | numeric `exclusiveMinimum` / `exclusiveMaximum` |
| `$schema` on a nested subschema | stripped — only the document root declares a dialect |

Today's emitted schema bodies contain **none** of those constructs, so the conversion is a no-op on them and only the dialect declaration changes. The single `$ref` we do emit (`search-messages`'s `.properties.dateTo` → `#/properties/dateFrom`) is a relative pointer valid in both dialects and is passed through untouched. The converter exists so the schemas stay honest if a new zod construct later introduces one of those keywords.

## Verification

**End-to-end MCP handshake against the rebuilt bundle** (`initialize` → `notifications/initialized` → `tools/list`):

| | `main` (2.10.11) | this branch (2.10.12) |
|---|---|---|
| tools advertised | 50 | **50** (unchanged) |
| `draft-07` occurrences in the `tools/list` payload | 100 | **0** |
| `https://json-schema.org/draft/2020-12/schema` declarations | 0 | **100** (50 input + 50 output) |
| total `$schema` keys | 100 | **100** — equal to the 2020-12 count, so no nested declarations |
| draft-07-only constructs on the wire | none | **none** |

Also booted the committed bundle **standalone in a temp dir with no `node_modules`** (what CI's `bundle-boots` job does): boots as v2.10.12, 50 tools, 0 draft-07, 100 × 2020-12.

**Tests**

- New `src/utils/jsonSchemaDialect.test.ts` — 27 unit tests covering the root stamp, nested-`$schema` stripping, `definitions`/`$defs` + ref rewriting, tuple `items` → `prefixItems` with `additionalItems` → `items`, a stray `additionalItems` being dropped rather than clobbering a non-tuple `items`, the `dependencies` split, both boolean-exclusive-bound collapses and their `false` counterparts, `tools/list` rewriting (including a tool with no `outputSchema`), non-`tools/list` messages returned byte-identical, and `withJsonSchema2020_12` returning the same instance while delegating the normalized message plus forwarded options.
- Extended `test/output-schema.test.ts` (the existing contract test that boots the real built server over stdio) with three assertions over what the server **actually advertises**: every schema declares 2020-12 and never mentions draft-07; no schema uses a draft-07-only construct; no subschema carries its own `$schema`.

**Gates** — `pnpm run lint` (0 errors), `pnpm run typecheck`, `pnpm test` (485 passed, 33 files), `pnpm run format:check`, plus the `output-schema` contract suite (7 passed) and `sync-plugin-version --check` / `sync-skills --check`. Bundle rebuilt and committed.

## Docs

- **README.md** — added a Troubleshooting entry for the exact error string, pointing at 2.10.12 and #147.
- Reviewed and unchanged: the Tool Reference (no tool, parameter, or output shape changed), `CLAUDE.md`, `docs/IMAP-SETUP.md`, `docs/NODE-RUNTIME-AND-TCC-PERMISSIONS.md`, `skills/` — none document the advertised schema dialect.


sweetrb:
## Follow-up: the converter is now position-aware

Pushed `b27bc59`. Fixes a **latent** defect in the converter this PR introduces — no version bump, it rides in the 2.10.12 release the PR already carries.

### The bug

`convertNode` recursed uniformly and then switched on **every key it met**. But a `properties` map's keys are caller-chosen **tool parameter names**, not schema keywords, so the keyword rewrites fired on user data:

| A tool declaring a parameter named… | shipped on the wire as |
|---|---|
| `definitions` | renamed to `$defs` |
| `$schema` | **silently deleted** — while `required` still listed it, so no input could satisfy the schema |
| `dependencies` | restructured into `dependentRequired`/`dependentSchemas` |
| `additionalItems` | **dropped entirely** |

Same class of error one level down: `enum` / `const` / `default` / `examples` hold instance **data**, so recursing into them rewrote a caller's literal values as if they were schema keywords — a `default` of `{ definitions: 1, $schema: "x" }` came out mangled.

### The fix

Keyword rewrites now apply **only at schema positions**:

- `SCHEMA_MAP_KEYWORDS` (`properties`, `patternProperties`, `$defs`, `dependentSchemas`) go through a new `convertSchemaMap`, which converts only the **values** and leaves the caller's names untouched.
- `DATA_KEYWORDS` (`enum`, `const`, `default`, `examples`, `required`, `dependentRequired`) pass through verbatim.
- `definitions` keeps its own case — it is a real keyword at a schema position — and its members now go through the map converter too, so a definition *named* like a keyword survives.
- The `$ref` case gains a comment recording a deliberate limitation: only the exact root prefix `#/definitions/` is rewritten, because `#/properties/definitions/...` could legitimately address a property *named* `definitions` and the two are indistinguishable without resolving the pointer. The SDK's converter emits neither construct.

### Why it was latent, not live

Verified against the current tool surface: **no server hits a corrupting name today.** apple-notes-mcp's `get-checklist-state` has an output property named `items`, which happened to land in a safe branch. Fixing it before the release rather than after.

### Evidence

Real MCP handshake against the rebuilt bundle — the advertised surface is **byte-for-byte identical** to the pre-patch payload:

```
toolCount: 50                          (unchanged)
inputSchemasDeclaring2020_12: 50/50
outputSchemasDeclaring2020_12: 50/50
draft07Substrings: 0
tools/list sha256: a769f704fb7c977404c9c39d3d35d565452af6cf92d492429e091bad32adbdc2
                   ^ identical before and after this commit
```

**Tests: 485 → 493 (+8).** Six of the eight fail against the pre-fix converter. The other two pass on both sides *by design* — they are non-regression guards proving the keyword behavior still works (a real `definitions` block still becomes `$defs` with its `#/definitions/X` `$ref` repointed) and that a `get-checklist-state`-shaped `outputSchema` round-trips unchanged apart from the root dialect.

Gates: `lint` / `typecheck` / `test` (493) / `format:check` all green, plus `test/output-schema.test.ts` (7) against the rebuilt bundle. Committed `build/index.js` verified byte-identical to a fresh `pnpm run build`.

Issue #31, opened 2026-06-17
Batch operations are N× full-tree-walk osascript spawns (slow at scale)


sweetrb:
## Summary
The batch tools fan out to N separate osascript processes, each doing a full mailbox-tree walk:
- `batchDeleteMessages` (~1656), `batchMoveMessages` (~1679), `batchMarkAsRead/Unread` (~1697/~1709), `batchFlagMessages/Unflag` (~1725+) each loop over the ids and call the single-id method.
- Each single-id method (`deleteMessage`, `moveMessageInternal`, `markAsRead`, …) spawns its own osascript and walks `accounts → mailboxes of acct → messages whose id is …` to locate one message.

So a 100-id batch = 100 processes, each a full breadth-first tree scan with a `whose id` query per mailbox, all serialized through the gate.

## Impact
Minutes-to-hours for a full batch on accounts with large mailboxes; the serial gate means nothing else runs meanwhile. The 100-id cap bounds but does not fix it.

## Fix
Replace the per-id fan-out with a single AppleScript that accepts the whole id list, walks the tree once, performs the action per matched id, and returns a per-id result vector; parse into `BatchOperationResult[]`. (For moves, resolve the destination once.) This turns N tree-walks + N process spawns into one.

_Severity: High · Effort: M · from the 2026-06-17 stability/perf audit._

PR #154, opened 2026-08-13
feat(batch): explicit source scope + surface per-id errors (follow-up to #153)

### LINKED ISSUE #152: Batch move/delete mis-targets messages after a mutation — list-messages IDs appear unstable across write ops (2.10.12)
First — 2.10.12 fixed the schema issue completely, reads are perfect. Thank you for the fast turnaround. Reporting a separate **write-path** problem I hit right after.

**Environment:** Claude Desktop (macOS), `npx -y apple-mail-mcp@2.10.12`, iCloud IMAP account.

**What I observed (in order):**

(1) `list-messages(INBOX)` returned numeric ids.
(2) `batch-move-messages` of 22 of those ids to a folder ("AI Stuff") → `success:22`, all correct.
(3) Re-listed INBOX to refresh ids.
(4) `batch-delete-messages` of 24 fresh ids from that new listing → reported `success:24`, but two of the intended ids were still in INBOX afterward, while two adjacent messages that were NOT in the id list got deleted instead.
(5) `batch-move-messages` of 2 fresh ids (Deleted Messages → INBOX) → reported `success:2`, but ~7 other INBOX messages moved to Trash as collateral.

**Observation:** the numeric ids returned by `list-messages` don't appear to be stable identifiers across calls once the mailbox is mutated — move/delete seems to resolve to a position/sequence that shifts as messages move, so operations land on neighbors. (IMAP sequence-number vs UID handling?) The reported `success` counts match the number of ids passed, not what actually happened, so they can't be used to verify the result.

**Impact:** single ops immediately after a list may be OK, but batch mutation is currently unsafe for automated workflows. Nothing was lost — everything was recoverable in Trash. Sibling servers presumably share the code path. Happy to send exact call/response logs if useful.

sweetrb:
Follow-up to #153 (which fixed #152 while this branch was in flight). **Rebased onto it — this PR no longer contains a competing implementation of the scoping fix.** What remains are the two gaps that leave #153's fix unreachable or unexplained in practice.

## History of this PR

I opened this having independently diagnosed and fixed #152. #153 landed first with the same diagnosis and a better-structured implementation (grouping ids per source mailbox, so each group opens one mailbox once, rather than my per-id scope lists). I merged `main` in and **took #153's implementation wholesale** for `appleMailManager.ts` and its `batchScope.test.ts`. Keeping only what is additive is the right outcome; the duplicate work is discarded.

For the record, my independent read-only reproduction agreed with #153's on every point — including the mechanism, which is worth restating because it is not the reporter's hypothesis.

## Correcting the reporter's hypothesis

The report suggested IMAP sequence-number vs UID instability. That is the wrong layer. `imap:` ids encode account + mailbox path + UID and are applied with a single `UID STORE`/`UID MOVE` against that one mailbox — they were never affected. The defect was confined to **numeric AppleScript ids**, and it was not instability over time: the same id resolves differently depending on *which mailbox you look in*, at the same instant. The instinct that ids were being reinterpreted between the list and the mutation was right; the layer was one up.

<details>
<summary>Read-only reproduction (no mail was mutated)</summary>

`listMailboxes` iterates `mailboxes` unsorted, so its output *is* the walk order the batch used:

```
accounts: [iCloud, robert.b.sweet@gmail.com, rob@superiortech.io]

rob@superiortech.io, in walk order:
   1. INBOX     …     5. [Gmail]/All Mail     …     12. Sales Spam
```

One numeric id, reachable from two mailboxes:

```
RFC Message-ID: <f1366f68-f3e1-4570-91bc-6c1fad130398@pivotpublic.com>
Mail.app numeric id: 79345
  - "Sales Spam"        IMAP UID 9      <- walk position 12
  - "[Gmail]/All Mail"  IMAP UID 90259  <- walk position  5   *** REACHED FIRST ***
```

A batch targeting the `Sales Spam` copy operated on the `All Mail` copy. For Gmail those are different operations: deleting the `All Mail` copy trashes the mail; deleting the `INBOX` copy only removes the Inbox label — which is exactly "reported deleted, still in INBOX".

**One thing I could not reproduce, stated plainly:** two *different* emails sharing a numeric id. On this machine ids come from a single store-wide counter, strictly monotonic with arrival time across *both* accounts (`79345`→`79375`→`79408`→`79440`→`79442`→`79464` tracks arrival order across two different accounts). So what I demonstrated is one message reachable under one id from several mailboxes — aliasing, not collision. #153's independent observation (id 75816 in `Important`, `All Mail` and `INBOX`) is the same phenomenon. The fix is correct for both, and #153's `idLocationIndex` comment about per-mailbox uniqueness is left as-is rather than rewritten from one machine's evidence.

I also could not reproduce the reporter's "two messages that were NOT in the id list had been deleted". The aliasing mechanism explains mis-targeting *within* a message's own alias set; it does not by itself explain touching an unrelated message. Flagging that as genuinely unexplained rather than papering over it.

</details>

### 1. The refusal message never reached the caller

#153 added a real diagnostic — an ambiguous id fails with every candidate mailbox named. But `hybridBatchCounts` in `src/index.ts` discards the AppleScript path's per-id `error` strings and keeps only a count:

```ts
const res = appleFn(numericIds);
const s = res.filter((r) => r.success).length;
success += s;
fail += res.length - s;     // <- r.error dropped on the floor
```

So a user whose batch was correctly refused sees `Deleted 22 message(s), 2 failed` and nothing more — no way to learn which ids were ambiguous, or that naming the mailbox would fix it. The carefully written message was unreachable.

Per-id reasons now surface in the text response and in an `errors` array on the structured result. `errors` is **declared in the batch output schema** so the failure channel is part of the advertised contract — clients can rely on the shape and it appears in generated types.

(An earlier revision of this description justified declaring it by claiming a raw zod shape compiles to `additionalProperties: false`, so an undeclared key would be rejected client-side with `-32602`. That is **false for this server** and has been retracted: `registerTool` wraps every outputSchema in `z.object(shape).passthrough()`, and booting the built server confirms all six batch tools advertise `"additionalProperties": true` — an undeclared key would have been carried through, not rejected. Declaring `errors` is still right; the old reason was not.)

### 2. The id→mailbox binding is per-process, with no way to supply it

#153 resolves ids against the `idLocationIndex` that list/search populates. That is in-memory, per-server state. A client that reconnects, restarts the server, or replays a stored list of ids has an **empty** index — so every id takes the whole-tree path, and on a label store (where one message answers to the same id in `INBOX`, `Important` and `All Mail`) is then refused as ambiguous. The fix is correct but the id becomes unusable, with the only remedy being "re-list the mailbox first", which a batch caller may not be positioned to do.

Adds `sourceMailbox` / `sourceAccount` to all six batch tools — the mailbox the numeric ids came from. It keeps those ids on the scoped path, and overrides a stale remembered location. For `batch-move-messages` these name the **source**, distinct from `mailbox` (the destination). `imap:` ids ignore both.

Also adds `AppleMailManager.noteMessageLocation(id, account, mailbox)` for callers holding ids obtained outside the process.

## Tests

New `src/services/appleMailManager.batchScopeExplicit.test.ts` (6 tests, mocked AppleScript layer — no live Mail.app). It asserts the behaviour that distinguishes this change: with a **cold** index and an explicit scope, the generated script targets the named mailbox and the whole-tree walk (`repeat with acct in accounts`) is **not generated at all**; without a scope and a cold index, it is. Plus: explicit scope beats a conflicting remembered location, `noteMessageLocation` puts an id on the scoped path, all six operations honour the scope, and the batch stays a single `osascript` invocation (#31).

**Against pre-#153 code**, the original version of these tests failed 8/10, the key one being the reported bug itself:

```
FAIL > operates on the INTENDED mailbox when an id exists in several mailboxes
  [ { account: "rob@superiortech.io",
-     mailbox: "Sales Spam",
+     mailbox: "[Gmail]/All Mail",     // wrong copy mutated
      id: "79345" } ]
FAIL > refuses (does not guess) when the scope is unknown and the id is ambiguous
  expected true to be false            // old code reported success
```

Against **current main**, the new tests fail to compile — `batchDeleteMessages` has no scope parameter there. That is the honest characterisation: this half is additive, not a second bug fix.

## Not covered

- No live-mail verification of the mutation path. The reproduction is strictly read-only by design; regression tests run against the mocked AppleScript layer.
- The `imap:` route is untouched — already mailbox-scoped by construction.
- **Sibling repos checked, none affected.** `apple-notes-mcp`'s `batchDeleteNotes`/`batchMoveNotes` address `note id <x-coredata://…>` directly — a globally unique URI, no tree walk, no first-match selection; its only account/folder walks are read-only stat aggregations. `apple-numbers-mcp` has no id resolution (path + sheet/table/cell coordinates). `apple-photos-mcp` passes osxphotos UUIDs straight through.

## Gates

`lint` (0 errors), `typecheck`, `test` (524/524), `format:check`, and the integration `output-schema` contract (8/8). Bundle rebuilt and committed; a fresh `pnpm run build` produces no diff. Version bumped to **2.10.16** with a real `## [2.10.15]`-style heading and an empty `## [Unreleased]`.


sweetrb:
Review follow-ups pushed as `f6d8e61`. Version stays **2.10.16** (still absent from npm; latest published is 2.10.15).

### 1. `sourceMailbox` without `sourceAccount` was a silent no-op

`runBatchOperation` only honored a scope when **both** fields were present, so a lone `sourceMailbox` was discarded, the ids fell back to the whole-tree path, and a label-store id was then refused as ambiguous — the exact failure the parameter exists to prevent. It was also inconsistent with the rest of the server, where an omitted `account` means *the default account* (`resolveAccount`), not *ignore your other argument*.

New `AppleMailManager.resolveBatchScope()` returns one of three outcomes:

- no `mailbox` → no scope (index / ambiguity-checked path, unchanged);
- `mailbox` + `account` → scoped, unchanged;
- `mailbox` alone → scoped to `resolveAccount()`'s default account; and if no account can be resolved (Mail returned none, or the transport failed) **every id in the batch fails** with an error naming the mailbox and asking for an explicit `sourceAccount`.

The safety property is absolute: a scope the server cannot honor is never quietly downgraded to the scan-and-guess walk. Three new tests in `appleMailManager.batchScopeExplicit.test.ts` cover the scoped-path emission, the `APPLE_MAIL_MCP_DEFAULT_ACCOUNT` override, and the fail-closed branch (asserting **no** AppleScript is generated at all).

Observed failures before the fix:
```
× scopes on sourceMailbox ALONE …
  expected [ { id: '79345', …(2) } ] to deeply equal [ { id: '79345', success: true } ]
× honors the default-account override …
  expected '\n tell application "Mail" …' to contain 'robert.b.sweet@gmail.com'
× FAILS the ids when no account can be resolved …
  expected 'No result returned' to contain 'Sales Spam'
```

### 2. The `additionalProperties` rationale was false — corrected in code and CHANGELOG

Verified by booting the committed bundle over stdio and dumping the advertised schema:

```json
{"$schema":"https://json-schema.org/draft/2020-12/schema","type":"object",
 "properties":{"ok":…,"success":…,"failed":…,"mailbox":…,
 "errors":{"type":"array","items":{"type":"string"}},
 "errorsTruncated":{"type":"boolean"}},
 "additionalProperties":true}
```

`additionalProperties` is **`true`**, not `false` — `registerTool` wraps every `outputSchema` in `z.object(shape).passthrough()` (its own docblock says so, and `test/output-schema.test.ts` asserts it). So an undeclared key would have been carried through, not rejected with `-32602`. Declaring `errors` is still correct — it makes the failure channel part of the advertised contract — so only the justification changed, in both the comment and the 2.10.16 entry. Leaving it would have shipped a superseded belief about this codebase into the GitHub Release notes.

### 3. Coverage for the headline change — and two defects it exposed

`hybridBatchCounts` / `formatBatchErrors` / the response shaping moved to `src/tools/batchResults.ts`. `src/index.ts` opens a stdio transport at import, so nothing in it can be imported by a test; extracting is what makes this half testable, and it collapses six drifting copies of the same three-branch block into one path. New `src/tools/batchResults.test.ts` (12 tests) covers per-id reasons surviving into both channels, the ambiguous-id refusal reaching the caller verbatim, and the dedupe/cap.

Two defects fixed:

- **all-fail branch** called `errorResponse(msg)` with no `structured`, so `errors` appeared in the text and *never* in `structuredContent` — for the one outcome a caller most needs to act on. All three branches now carry the same structured payload. Observed failure: `expected undefined to match object { ok: false, success: +0, failed: 3 }`.
- **uncapped structured `errors`** while the text was capped at 5 distinct: a 100-id all-ambiguous batch emitted 100 long strings, each naming every mailbox holding that id. Now deduped and capped at `MAX_STRUCTURED_BATCH_ERRORS = 20`, with `errorsTruncated: true` when the cap bites so a short list is never read as the whole story. (A partial batch's structured `errors` also no longer repeats one reason 22 times.) Observed failures: `expected undefined to deeply equal [ 'Message not found' ]` and `Target cannot be null or undefined`.

### 4. Inherited doc error from #153

The docblock asserted `mailboxes of account` "yields INBOX late", so INBOX ids hit the `Important` copies. Checked with `list-mailboxes` on the reporting account (2026-08-13): **INBOX 1, `[Gmail]/All Mail` 5, `[Gmail]/Important` 9, `Sales Spam` 12** — and `listMailboxes()` does not sort, so that is raw walk order. The real mechanism is first-match-wins handing the operation to whichever alias the walk reaches *first*, so **any** late-walked source mailbox loses — which is the `Sales Spam` vs `All Mail` case this PR reproduces. INBOX happened to be first there, not last. Corrected in `runBatchOperation`, in `findMessageScript`, in the `batchScope` test header, and called out in the 2.10.16 CHANGELOG (2.10.15's shipped entry is left as the historical record it is).

### Docs / gates

`sourceMailbox`/`sourceAccount` docs updated in the README Tool Reference and in both zod `.describe()` strings to say `sourceMailbox` scopes on its own and `sourceAccount` alone pins nothing; the two `Returns:` lines that promised "an error per failed id" now say "the distinct reasons for any failures".

- `pnpm run lint` — 0 errors (10 pre-existing `no-explicit-any` warnings)
- `pnpm run typecheck` — clean
- `pnpm test` — **539 passed / 39 files**
- `pnpm run format:check` — clean
- `vitest --config vitest.integration.config.ts test/output-schema.test.ts` — **8 passed** (the contract CI runs)
- `build/index.js` + `build/cli.js` rebuilt and committed; a fresh `pnpm run build` leaves **zero** git diff — `index.js` `sha256:8c0222602bba59fd68334b79f46b18d86d9cf83d5363211b04c25a219483c78d`, `cli.js` `sha256:cebf87819390757db813219f6551ce12ffa88d6e9c24fbcf9895d348508e9200`

`[Unreleased]` is present and empty. Not merged.

Issue #155, opened 2026-08-13
Unexplained: batch delete/move affected messages whose ids were never passed (#152 residual)


sweetrb:
Splitting an unexplained residual out of #152 so it doesn't get lost now that the parent is closed.

## What was explained, and what wasn't

#152 reported three symptoms from a `batch-delete-messages` / `batch-move-messages` sequence on 2.10.12. Two are explained and fixed in 2.10.15 (#153):

- intended messages surviving the operation, and
- a `success` count that matched the input rather than reality.

Root cause was by-id resolution walking every account × mailbox and operating on `item 1` of the first match. On a Gmail-backed store one message is reachable through several mailboxes (`INBOX`, `[Gmail]/All Mail`, `[Gmail]/Important`), so `whose id is N` matches in all of them and the walk order decided which copy got hit — which matters because deleting the All Mail copy trashes the message while deleting the INBOX copy only removes a label.

**The third symptom is not explained by that mechanism**, and I don't want it absorbed into a fix that doesn't actually address it. From #152 step (4):

> two of the intended ids were still in INBOX afterward, while **two adjacent messages that were NOT in the id list got deleted instead**

Alias resolution explains operating on the wrong *copy of a listed message*. It does not explain operating on a message whose id was never passed. Step (5) has the same shape — a 2-id move producing ~7 messages moved as collateral.

## Candidate explanations, none verified

- `delete` on a Gmail label-mailbox having store-wide effects beyond the addressed message;
- Mail.app re-indexing numeric ids mid-batch, so a later id in the same script resolves to a message that arrived at that id only after an earlier delete;
- the reporter's own reading of "adjacent" being a different message than they thought (worth ruling out, not assuming).

The second is the one I'd chase first: the batch runs as a single AppleScript, and nothing in the old implementation re-validated an id between operations within that script.

## Why this is open rather than closed

I could not reproduce it. Both enabled accounts on the machine I investigated from route through the IMAP path and return `imap:` ids, which encode account+mailbox+UID and are unaffected by this class of defect — so the numeric-id path is barely exercised there. I'd rather leave a known-unexplained data-loss-shaped symptom tracked than quietly assume the parent fix covered it.

**If you hit anything like this on 2.10.15 or later, please comment here** — particularly the ids passed, the ids that actually changed state, and whether the account is Gmail/iCloud/other. That last detail is what decides between the candidates above.

[...]

PR #157, opened 2026-08-13
feat(forensics): make destructive ops report what they actually did (#155)

sweetrb:
Refs #155. **Instrumentation, not a fix** — #155 is still open and still unreproduced. This makes it *self-reporting* if it happens again.

## Why

#152 reported three symptoms. Two were explained and fixed in 2.10.15/2.10.16: by-id resolution walked every mailbox and acted on whichever alias came first. That mechanism explains operating on the **wrong copy of a listed message**. It does not explain operating on a message **whose id was never passed** — the residual now tracked as #155.

It is unreproducible today for one reason: **nothing recorded what a delete actually did.** Success meant "the AppleScript did not throw", never "a message was observed to go away". So this adds what a diagnostician would need after the fact to explain "these two messages vanished and I never asked for them".

### 1. `countDelta` — ALWAYS ON, zero extra `osascript` calls

`delete-message`, `move-message`, `batch-delete-messages`, `batch-move-messages` count the affected **source** mailbox immediately before and immediately after the mutation and compare observed with expected. Both counts are taken **inside the AppleScript the operation was already running** — count before the loop, count after — so the single-invocation property from #31 is unchanged. Declared in each tool's output schema.

`status` is four-valued, and **only `over` warns**:

| `status` | Meaning | Warns? |
|---|---|---|
| `match` | as many left as were operated on | no |
| `over` | **more** left than were operated on — the #155 signature | **yes** |
| `under` | fewer left than expected | no |
| `unknown` | Mail would not report a count | no |

That asymmetry is the design, not laziness. New mail arriving mid-operation *raises* the after-count, which can only bias a reading toward `under` — so `over` cannot be manufactured by ordinary traffic. `under`, by contrast, is routine: an account that flags deletions rather than removing them leaves the count unmoved while the delete genuinely succeeded. Warning there would fire on ordinary Gmail deletes, and a warning that cries wolf is ignored exactly when it matters. `under` is reported in the field with a `note` and never warned about.

The Gmail expectation was worked out rather than assumed: deleting the `INBOX` copy drops the `\Inbox` label and deleting the `[Gmail]/All Mail` copy trashes the message — different operations, but the **source** mailbox loses exactly one entry per successful id either way. The two cases that are genuinely unpredictable are said plainly instead of guessed: a move whose destination **is** the source expects `0` (with a `note`), and an unreadable count yields `status: "unknown"` with null counts. `imap:` ids are not reconciled at all — an IMAP UID names one message in one mailbox, so this defect class cannot occur there, and no `countDelta` is invented for them.

### 2. `APPLE_MAIL_MCP_AUDIT_LOG` — opt-in NDJSON forensic log

Off by default and generating **no extra Apple Events** when off (the pre-image fragment is only emitted into the script when the log is enabled). One record per destructive op: timestamp, tool, server version, arguments, the **pre-image** of every resolved message (account, mailbox, numeric id, RFC Message-ID, `date received`), the per-id outcome, and the reconciliation.

### 3. Collateral identification — opt-in, gated on (2)

`(numeric id, Message-ID)` pairs snapshotted before and after and diffed, so the log **names** what disappeared, including messages the caller never listed (`unrequested`). That is the evidence #155 is missing. O(mailbox size), so bounded by `APPLE_MAIL_MCP_AUDIT_SNAPSHOT_MAX` (default 2000; `0` disables) — and **when the bound bites the skip is recorded with its reason**, because a silently skipped snapshot reads as "nothing collateral happened".

## Privacy

Default logs what **identifies** a message (Message-ID, date, mailbox, account, numeric id) and nothing about what it says. **Subjects are behind a second, separate opt-in** (`APPLE_MAIL_MCP_AUDIT_SUBJECTS=1`) — a subject line is frequently the whole sensitive payload, and it is not needed to diagnose #155. **Bodies are never logged, under any setting.** No default log path, so enabling it is deliberate; the README states that the file grows unbounded and is never rotated.

**stdout is untouched** — it is the JSON-RPC transport. Records go to the configured file, diagnostics to stderr, and a failed write is reported and swallowed rather than failing a mutation that already happened. A test asserts stdout is never written.

## Acceptance: would this have caught #155?

Mocked scenario with the exact reported signature — a 2-id delete where **4** messages leave, **2** of which were never requested:

```
=== TOOL RESULT (what the caller sees) ===
text:
  Successfully deleted 2 message(s)

  ⚠️ Effect mismatch in "INBOX" in account "me@example.com": 4 message(s) left the
  mailbox but only 2 were operated on (count 5 → 1). 2 message(s) are unaccounted
  for. This is the signature of https://github.com/sweetrb/apple-mail-mcp/issues/155
  — please report it there, and set APPLE_MAIL_MCP_AUDIT_LOG=/path/to/audit.ndjson
  to capture which messages disappeared.

structuredContent.countDelta:
[ { "account": "me@example.com", "mailbox": "INBOX",
    "before": 5, "after": 1, "expected": 2, "observed": 4, "status": "over" } ]

=== AUDIT LOG (collateral section of the NDJSON record) ===
"collateral": [ { "account": "me@example.com", "mailbox": "INBOX", "snapshot": "ok",
  "disappeared": [ {"id":"75811","messageId":"a@example.com"},
                   {"id":"75812","messageId":"b@example.com"},
                   {"id":"75814","messageId":"d@example.com"},
                   {"id":"75815","messageId":"e@example.com"} ],
  "unrequested": [ {"id":"75814","messageId":"d@example.com"},
                   {"id":"75815","messageId":"e@example.com"} ],
  "appeared": [] } ]
```

The warning fires **and** the log names the two unrequested messages.

## Tests

18 new tests through the mocked `executeAppleScript` layer. The mock is a **simulator**, not a canned string: it parses the generated script (id list, snapshot ceiling), mutates a fake mailbox, can inject the #155 defect, and emits a pre-image/snapshot **only when the generated script actually asked for one** — so instrumentation that is silently removed makes the tests fail rather than pass vacuously.

Covers: mismatch detected; no false alarm on the normal case; `under` does not warn; destination-is-source expects 0; pre-image + per-id outcome written; subjects only behind the second opt-in; **audit path never writes to stdout**; an unwritable log path does not fail the mutation; collateral names an unrequested deletion; the skip-above-ceiling path records the skip; the non-destructive batches are not instrumented; and one asserting the reconciliation names the mailbox the script actually opened.

## Gates

`pnpm run lint` (0 errors, 10 pre-existing warnings) · `typecheck` · `test` 557/557 · `format:check` · integration `test/output-schema.test.ts` 8/8. `build/index.js` rebuilt and committed; a fresh rebuild leaves zero git diff (`b6fe97f9c5b83e513e3b31e5d7c333ac4fdaaae563da1c40ad9d211c30303437`). `build/cli.js` unchanged.

Version bumped 2.10.16 → **2.10.17** with a real CHANGELOG heading; `## [Unreleased]` left empty. No sibling repo touched.

https://claude.ai/code/session_01VUsvik7hHLhhxQ6MHTXeMx


sweetrb:
All five review findings fixed in `305018d`. Version stays **2.10.17** (not yet on npm; `latest` is 2.10.16), `## [Unreleased]` is present and empty.

The two that mattered most both attack the same property — that the always-on warning is worth reading. Every new test below was **verified failing against the pre-fix code** first.

---

### 1. A duplicated id fabricated an `over` warning on a *correct* delete

`expected` was derived through `byId`, keyed by the id string, so `['75811','75811']` counted the id twice while the mailbox only ever lost one message.

**Decision at the API level: a batch is a SET.** A repeated id names the same message, and Mail can only delete it once — so:

- ids are deduped on the **numeric value actually sent to AppleScript** (`"75811"` and `" 75811"` collapse too — same target);
- the batch returns **one result per distinct id**, so `success` counts messages rather than list positions (also deduped in `hybridBatchCounts`, so the rule holds for `imap:` ids as well);
- `expected` is now read from the record stream's own **1-based positions**, which are unique by construction, instead of through an id-keyed map.

Documented in the README (batch preamble + the `countDelta` honesty rules) and the CHANGELOG.

**Tests:** `a repeated id names ONE message` — (a) no warning, one operand in the generated `_gids`, one result; (b) dedupe on the numeric value; (c) dedupe does **not** blunt detection — the same duplicated call with two unrequested messages vanishing still warns.

Observed before the fix, on a delete that did exactly the right thing:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+ [
+   "⚠️ Effect mismatch in \"INBOX\" in account \"me@example.com\": 1 message(s) left the
+    mailbox but only 0 were operated on (count 5 → 4). 1 message(s) are unaccounted for.
+    This is the signature of .../issues/155 — please report it there, ..."
+ ]
```

### 2. The "cannot be manufactured" claim was stronger than the evidence

Correct in all three places (CHANGELOG, README "Auditing destructive operations", the `countDeltaWarning` doc comment) — and the reasoning is corrected in strength, not deleted:

- **What `over` establishes:** more messages left the source mailbox across the operation's window than the operation accounted for. The data-loss direction, and the #155 signature.
- **What it does not:** that *this server* removed them. Any concurrent **departure** reads identically — a Mail rule firing mid-batch, a server-side filter, another client, an IMAP expunge landing between the counts. That is now named as the benign cause to rule out **first**.
- The asymmetry argument survives for exactly what it covers: concurrent **arrivals** cannot produce `over`, since they raise the after-count and bias toward `under`. So `over` is the interesting direction and a strong signal — **not a proof**.

The warning text itself now says so, so the caveat reaches the person who sees the warning and not only the person who reads the docs:

> ⚠️ Effect mismatch in "INBOX" in account "me@example.com": 4 message(s) left the mailbox but only 2 were operated on (count 5 → 1). 2 message(s) are unaccounted for. **Anything else removing mail from this mailbox at the same moment — a Mail rule, a server-side filter, another client, an IMAP expunge — reads the same way, so rule that out first. If nothing else was touching it,** this is the signature of …/issues/155 — please report it there, and set `APPLE_MAIL_MCP_AUDIT_LOG=…` to capture which messages disappeared.

### 3. Externally controlled bytes in the delimited stream

The Message-ID (always) and the subject (`APPLE_MAIL_MCP_AUDIT_SUBJECTS`) are written by whoever sent the mail, so a crafted Message-ID carrying `RS` + a forged `RECON` tag could inject a reconciliation record — fabricating the exact evidence the feature exists to produce, and only ever while someone is chasing a real incident.

Stripped **at the source**, in the same AppleScript that reads the values (one `text items` pass — AppleScript takes a *list* of delimiters when splitting and the first when joining), covering the pre-image Message-ID / date / subject, every snapshot Message-ID, and the runtime account + mailbox names on the unlocated paths. Values interpolated from TypeScript (account/mailbox literals) go through `stripStreamDelimiters`. Replaced with `U+FFFD` rather than dropped, so an altered value shows that it was altered.

Verified against **real `osascript`** (pure AppleScript, no Mail.app), with the exact fragment the manager generates:

```
input mid : evil<RS><GS>RECON<GS><US>acct
output    : evil��RECON��acct
snapshot  : 75811<GS>P<GS>a@example.com<GS>I<GS>75812<GS>P<GS>b�x@evil
```

— the record's own `<GS>P<GS>` / `<GS>I<GS>` framing intact, the injected `<US>` inside the value gone.

**Tests:** `a hostile Message-ID cannot forge the evidence` — a Message-ID embedding `RECORD_SEP` + a forged `RECON` claiming `5000 → 0`. Exactly one `countDelta` (the real one, `match`), no warning, and the logged Message-ID contains no delimiter but does contain `U+FFFD`. The simulator only strips if the generated script **asked** it to, so the test cannot pass vacuously on a defence that was removed. Pre-fix, the forged record landed:

```
- "before": 2,      + "before": 5000,
- "expected": 1,    + "expected": 3,
- "observed": 1,    + "observed": 4999,
- "status": "match" + "status": "over"
```

### 4. `lastForensics` could outlive its operation

One rule now, no exceptions: `beginMutation()` clears it at the start of **every** message mutation — instrumented or not, batch or single (via `runBatchOperation` for all six batch tools, `findMessageScript` for all five single-message ones, and `moveMessage`, which builds its own script). So the only two answers are "the report for the call I just made" and `undefined`.

**Tests:** a `batchMarkAsRead` after an unconsumed `batchDeleteMessages`, and a `markAsRead` after an unconsumed `deleteMessage` — both must yield `undefined`. Pre-fix both returned the delete's `countDeltas`.

### 5. Literal NUL bytes made the file binary to ripgrep

All six sites (five new, one pre-existing) now write the separator as the escape `\u0000`, via the shared `groupKey()` and a new `snapshotKey()`. Confirmed: `rg -c groupKey src/services/appleMailManager.ts` → `5` (it previously answered `binary file matches` and refused to search). No NUL remains in the source or in the committed bundle.

---

### Gate

`lint` (0 errors, 10 pre-existing `any` warnings) · `typecheck` · `test` **564 passed / 40 files** · `format:check` clean · integration **output-schema contract 8/8**. Bundle rebuilt and committed; a fresh `pnpm run build` afterwards leaves **zero** git diff. (The live-Mail integration tests fail in this environment with `Permission denied … Automation` — no Apple Events access for the test runner — which is environmental and unrelated.)

### #155 acceptance, re-proved after the changes

2 ids passed, 4 messages leave:

```
countDelta: [{ account: "me@example.com", mailbox: "INBOX",
               before: 5, after: 1, expected: 2, observed: 4, status: "over" }]

warning: ⚠️ Effect mismatch in "INBOX" in account "me@example.com": 4 message(s) left the
mailbox but only 2 were operated on (count 5 → 1). 2 message(s) are unaccounted for.
Anything else removing mail from this mailbox at the same moment — a Mail rule, a
server-side filter, another client, an IMAP expunge — reads the same way, so rule that out
first. If nothing else was touching it, this is the signature of
https://github.com/sweetrb/apple-mail-mcp/issues/155 — please report it there, and set
APPLE_MAIL_MCP_AUDIT_LOG=/path/to/audit.ndjson to capture which messages disappeared.

collateral.unrequested: [ { id: "75814", messageId: "d@example.com" },
                          { id: "75815", messageId: "e@example.com" } ]
```

Not merged.

sweetrb:
## Final round — all four findings addressed

Worktree `feat/delete-forensics` @ `e3cd246`. Version held at **2.10.17**; `## [Unreleased]` present and empty.

The through-line held throughout: **the only value of this feature is that its evidence can be trusted.** Two of these four were the feature failing at exactly that.

---

### A — the self-move false alarm, and the comment that denied it

**Reproduced first.** 3 of 3 self-move probes fired the full #155 warning (batch, batch-of-N, single-message), while `appleMailManager.ts:3754` and a test *title* both asserted the opposite.

**Behaviour fixed, not the wording.** `expectedFor` may now return **`null`** = "not predictable"; null propagates through the per-mailbox merge (absorbing — one unpredictable contribution makes the total unpredictable) and is classified `unknown`, which never warns. `countDeltaWarning` also refuses on `expected === null` independently, so a future status change can't resurrect a warning with no expectation behind it.

That is the honest answer rather than a suppression: a self-move should remove nothing, but what Mail does to the count when a message is re-filed into the mailbox it already occupies is *unspecified*, so any number there is a guess. It is also what `auditLog.ts`'s own `unknown` doc already claimed the design was — the code had drifted from its own interface contract.

`CountDelta.expected` and the `countDelta` output schema are now nullable, exactly as `before`/`after`/`observed` already were. Comment, note text (`SELF_MOVE_NOTE`, shared by both paths so they can't diverge), test titles, CHANGELOG and README all rewritten to describe what the code does.

**Tests (4), all verified failing pre-fix:** the three self-move paths, plus a guard that the suppression does **not** blunt the detector for an over-effective move to a *different* mailbox.

### B — the fabricated collateral

`unrequested` was keyed on the raw numeric id. AppleScript renders a Mail id above 2^29 in scientific notation (`999999999` → `9.99999999E+8`) while the caller's list is decimal, so on a large-id mailbox **every requested message was reported as collateral** — the log naming innocent messages as destroyed, in precisely the scenario where someone is reading it to find out what was destroyed.

Both sides now go through `canonicalNumericId`, applied at the **parse boundary** (`parseSnapshot`), so the fix also makes the id in the report the decimal one the caller passed rather than `9.99999999E+8` — otherwise the record is unusable for correlation even when it's right.

**Tests (3), verified failing pre-fix:** a requested >2^29 id is not reported as collateral; a genuinely unrequested one still *is*, named in decimal; plus a guard on the simulator's own fidelity (that it renders large ids the way AppleScript does), so the other two can't quietly become vacuous.

### C — the overstated claim (claim fixed, guard kept)

Confirmed: `escapeForAppleScript` strips `[\x00-\x1f\x7f]` at `appleMailManager.ts:408`, so account/mailbox names could never have carried a delimiter in from TypeScript. The CHANGELOG now says what was actually vulnerable — the values read **inside** the script, after escaping no longer applies (Message-ID, `date received`, subject, runtime mailbox/account names, Mail's error text, the ambiguity candidate list) — and marks the TS-side stripping as the defence in depth it is, in its own bullet. The redundant guard stays.

### D — the invariant, everywhere

`stripStreamDelimiters`/`sanitizeFragment` applied to the `source mailbox "X" not found` record (:3588) and the runtime-built `_unames` ambiguity list (:3654) — **and** to the two `error:` records carrying text Mail itself composed, which are runtime-read values by the same argument. The invariant is now stated where the record tags are defined, since that is what the next emitter's author will read.

**Tests (2 behavioural + 1 structural), verified failing pre-fix.** The simulator was upgraded to check sanitisation **per variable** (`set _zParts to text items of <var>`) instead of once per script — so one emitter's fragment can no longer vouch for another's, and an injection test can't pass because some *other* value happened to be sanitised. Injection through Mail's error text and through the ambiguity list are now real end-to-end tests (forged `RECON` in, no fabricated reconciliation out, error still delivered with `U+FFFD` markers).

The "mailbox not found" case is honestly *not* a live hole (per finding C's own reasoning), so its test asserts the observable difference instead: the name now shows it was **altered** (`U+FFFD`) rather than being silently shortened into something that reads as a real mailbox name — which is what `DELIMITER_REPLACEMENT` says it is for.

---

## Acceptance properties, re-proved after the changes

**(1) The #155 signature is still caught.** 2 ids in, 4 messages out:

```
countDelta: {"account":"me@example.com","mailbox":"INBOX","before":5,"after":1,
             "expected":2,"observed":4,"status":"over"}
warnings (1):
⚠️ Effect mismatch in "INBOX" in account "me@example.com": 4 message(s) left the mailbox
but only 2 were operated on (count 5 → 1). 2 message(s) are unaccounted for. …
collateral: {"snapshot":"ok",
  "disappeared":[75811 a@, 75812 b@, 75814 d@, 75815 e@],
  "unrequested":[{"id":"75814","messageId":"d@example.com"},
                 {"id":"75815","messageId":"e@example.com"}]}
```

And the same acceptance on a **>2^29 mailbox** — the case that previously named both messages, including the requested one:

```
countDelta: {"before":3,"after":1,"expected":1,"observed":2,"status":"over"}   warnings: 1
collateral: {"disappeared":[{"id":"999999999",...},{"id":"2147483647",...}],
             "unrequested":[{"id":"2147483647","messageId":"big-c@example.com"}]}
```

**(2) False-alarm rate: 0/22** — now including all three self-move cases and both large-id cases. Before this change the same matrix was 3 warnings + 2 fabricated-collateral rows:

```
ok   | delete 1 id / 2 ids / all 5 / unscoped / mixed / absent   | match     warnings=0
ok   | delete a repeated id (and with whitespace)                | match     warnings=0
ok   | delete on a flag-only store                     status=under         warnings=0
ok   | single-message delete                           status=match         warnings=0
ok   | move 1 id / 3 ids to another mailbox            status=match         warnings=0
ok   | SELF-MOVE: batch, destination IS the source     status=unknown       warnings=0
ok   | SELF-MOVE: batch of 2, destination IS source    status=unknown       warnings=0
ok   | SELF-MOVE: single message, dest IS source       status=unknown       warnings=0
ok   | single-message move to another mailbox          status=match         warnings=0
ok   | audit log ON / +subjects / ceiling bites / snapshot off  match  warnings=0
ok   | delete a large (>2^29) id, audit log ON         status=match  falseCollateral=0
ok   | delete two large (>2^29) ids, audit log ON      status=match  falseCollateral=0
--------------------------------------------------------------------------------
warnings on a correct operation: 0/22      (was 3/22)
fabricated collateral entries:   0/22      (was 2/22)
```

## Gate

`lint` (0 errors, 10 pre-existing `any` warnings) · `typecheck` · `test` **573 passed / 40 files** · `format:check` clean · integration `test/output-schema.test.ts` **8/8**. Bundle rebuilt and committed; a fresh `pnpm run build` afterwards leaves **zero** git diff.

Every one of the 9 new tests was run against the pre-fix source and observed to fail — 8 via a reverted-fix mutation of `appleMailManager.ts`/`auditLog.ts`, and the simulator-fidelity guard by construction.

⚠️ `test/integration.test.ts` fails on this machine with `Permission denied. Grant automation access` — a TCC/Automation gap for the runner, unrelated to this diff (no live Mail.app was touched; all behavioural testing went through the mocked `executeAppleScript` layer).

Not merged.

Issue #156, opened 2026-08-13
Follow-ups from #154 review: lone sourceMailbox scopes to the default account, plus three smaller gaps


sweetrb:
Three non-blocking findings from the adversarial review of #154, split out so they don't vanish with the merged PR. None is a regression — #154's safety properties were verified intact (real per-mailbox binding, ambiguity refused, no false-ok path, no silent fallback to the old scan-and-guess).

## 1. A lone `sourceMailbox` resolves to the **default** account

`batch*Messages(ids, { mailbox: "X" })` without `sourceAccount` resolves the account via `resolveAccount()` — `APPLE_MAIL_MCP_DEFAULT_ACCOUNT` → Mail's default-send account → first enabled account.

So ids that actually came from a **non-default** account get scoped to the *default* account's same-named mailbox. If that id exists there as a different message, it is mutated silently — the #152 failure mode translated from cross-mailbox to cross-account. Note the unscoped path would have *refused* that case as ambiguous, so for this specific shape the scoped path is less safe than the fallback.

Not reachable on the machine this was reviewed from (numeric ids there come from a store-wide monotonic counter, so a given id exists once and a wrong-account scope yields `notfound` rather than a mis-target), but that is a property of one store, not a guarantee.

Options: verify the mailbox name is unique across accounts before accepting a lone `sourceMailbox`; or require `sourceAccount` whenever the name exists in more than one account; or confirm the resolved message's RFC `Message-ID` matches what the caller listed.

## 2. A whitespace-only `sourceMailbox` is silently discarded

`sourceMailbox: "   "` trims to empty and is treated as no-scope, falling back to the whole-tree ambiguity-checked path. Safe — it refuses rather than guesses — but silently ignoring a supplied argument is exactly the class #154 fixed for `sourceMailbox`-without-`sourceAccount`. It should be rejected outright.

## 3. `resolveAccount()` can return a **disabled** account

When every account is disabled, `resolveAccount()` falls back to `accounts[0].name` regardless of the `enabled` flag, so a lone `sourceMailbox` can scope to a disabled account. Pre-existing and shared — `runBatchOperation` never consulted `disabledAccountGuard` before #154 either — but it is now reachable through a new parameter, which raises its priority.

## 4. No test covers `index.ts`'s wiring of the new parameters

[...]

PR #159, opened 2026-08-14
fix(security): require atomic numeric batch scope

jjoanna2-debug:
Summary
- Require numeric batch callers to provide the complete `sourceAccount` + `sourceMailbox` pair.
- Reject whitespace-only source fields instead of silently falling back to an unscoped walk.
- Close the cross-account wrong-target path documented in open issue #156, while leaving self-scoped `imap:` ids unchanged.

Verification
- `pnpm run build`
- `pnpm test` — 40 files, 573 tests passed
- `pnpm run lint` — 0 errors; 10 pre-existing `no-explicit-any` warnings
- `pnpm run format:check`
- `git diff --check`
- Commit hook reran `tsc --noEmit` and `vitest run` — 40 files, 573 tests passed

Risk / Notes
- Shipped runtime and public docs changed; package version is 2.10.18 and plugin manifests/build were synchronized.
- This is intentionally fail-closed and compatibility-breaking for callers that supplied only `sourceMailbox`; the error tells them to provide both fields.
- Open for maintainer review; no live Mail.app integration was run in this environment. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.23",
+      "version": "2.10.24",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.23",
+      "version": "2.10.24",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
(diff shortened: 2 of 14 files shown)


sweetrb:
Merged as 2.10.24 — this closes items 1 and 2 of #156, which was open for exactly this.

Your framing is what settled it for me: a mailbox name alone isn't an identity, so resolving the missing account from mutable default-send state can cross an account boundary silently. That made the lone-`sourceMailbox` path *less safe than passing no scope at all*, since the unscoped path refuses the ambiguous case outright. Removing the `resolveAccount()` call from `resolveBatchScope` makes that unreachable rather than unlikely, which is the right kind of fix.

I considered the more ergonomic option (accept a lone `sourceMailbox` when the name is unique across accounts) and rejected it: uniqueness is a property of the account set at check time, and a mailbox that's unique when cached and ambiguous when used is precisely the bug this parameter exists to prevent. Reasoning recorded on #156 — it's easy to loosen later, hard to tighten once people depend on the loose form.

Rebased onto 2.10.23 and renumbered from 2.10.18 to 2.10.24 on merge.

PR #160, opened 2026-08-14
fix: refuse ambiguous message id lookups

jjoanna2-debug:
Summary
- Route reply and forward operations through the existing account-and-mailbox resolver so a numeric Mail.app ID cannot mutate the first mailbox encountered.
- Make unlocated content and raw-source reads proceed only when exactly one mailbox contains the ID.
- Add deterministic AppleScript-shape regression coverage and release the change as 2.10.19 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 41 files, 575 tests passed
- `pnpm run format:check`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 41 files, 575 tests passed
- Apple Mail integration tests: Not run; this PR has deterministic mocked AppleScript coverage only.

Risk / Notes
- This is a behavior-hardening change for shipped runtime code and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- An ID with multiple mailbox matches now returns no read result or mutation rather than guessing; callers should list/search the intended mailbox first.
- Open for maintainer review; no live Mail.app state was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.24",
+      "version": "2.10.25",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.24",
+      "version": "2.10.25",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.24",
+  "version": "2.10.25",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.10.24",
+  "version": "2.10.25",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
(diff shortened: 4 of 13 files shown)


sweetrb:
Good problem to pick — the read path really is the last place still doing scan-and-take-`item 1`, and the reply/forward consolidation onto `findMessageScript` is the right move. Two changes needed before this can merge.

**1. The read path swallows the ambiguity instead of reporting it.**

`findMessageScript`'s no-location branch already handles this case and returns an actionable error:

```
error:<AMBIGUOUS_ID_PREFIX>42 is present in more than one mailbox (Gmail/INBOX, Gmail/[Gmail]/All Mail, );
list or search that mailbox first so the operation targets the right copy
```

The new `getMessageContent` / `getRawSource` scripts collect `_hits` the same way but fall through to `return ""` when the count isn't 1 — which the callers turn into `null`, surfacing to the user as an indistinguishable "message not found". That's strictly worse than the status quo for the affected user: before, they got the wrong-copy content (bad but visible); now they get a phantom "doesn't exist" for a message that plainly does, with nothing pointing at the fix. Please emit the same `AMBIGUOUS_ID_PREFIX` error these two paths' callers can surface.

**2. Neither read consults the recorded location first.**

This is what makes the change safe to ship. `findMessageScript` tries `locationFor(id)` before falling back to the scan, so the ordinary `list-messages` → `get-message-content` flow resolves against the mailbox the id was actually listed from and never reaches the ambiguity check. The new read scripts skip `locationFor` entirely and go straight to the all-accounts walk — so on a Gmail-backed store, where one message is legitimately reachable through `INBOX`, `[Gmail]/All Mail` and `[Gmail]/Important`, `_hits` is 3 for a perfectly normal message and **every numeric-id read fails**, including immediately after a list that recorded the exact location.

That's a hard regression for Gmail users on the AppleScript path, and the tests won't catch it: `executeAppleScript` is mocked, so they assert on generated script text and never exercise a multi-hit store.

Taking the same location-first-then-ambiguity-checked shape `findMessageScript` already uses fixes both points at once — ideally by reusing it rather than open-coding the walk a third time.

The reply/forward half of this PR I have no concerns about; that part is a clean deduplication onto a resolver that already got this right.

jjoanna2-debug:
Addressed both points. Numeric reads now use the recorded account and mailbox first, then fall back to a full scan; zero matches return not-found and multiple matches return the same actionable ambiguity error instead of null or an arbitrary item. The MCP handler consumes that diagnostic. Added focused regression coverage for the resolver-generated read scripts and ambiguity propagation. Final head `8aeb7c745a978ea8242dc62f82e2b580fae37ba3`; local and live checks are green.

sweetrb:
Merged and shipped in **2.10.25**. Thank you — routing numeric reads through the same location-aware resolver the batch mutations already used is the right unification; the read paths having their own scan-and-guess was a real inconsistency, and refusing a multi-mailbox match rather than picking one is the correct call.

One fix pushed to your branch before merge: the refusal was signalled by `return "error:…"` and detected with `output.startsWith("error:")` — but the success payload of the *same* script leads with `msgSubject`, which is whatever the sender wrote. So any message with a subject starting `error:` was reported as a failed lookup, became unreadable through `get-message`, and had the rest of its own payload stored as the error string. `getRawSource` shared the sentinel.

Switched both read paths to the GS-wrapped `\x1dERR\x1d` marker, which is the convention the file already uses for `CONTENT`/`MSGID`/`HTML` precisely because payload text cannot contain it. The bare `error:` prefix is left alone on the script paths whose output isn't caller-controlled. Two regression tests; reverting the marker fails 3.

PR #161, opened 2026-08-14
fix: prevent attachment overwrites

jjoanna2-debug:
Summary
- Refuse `save-attachment` when the destination already exists, including regular files and symlinks.
- Stage AppleScript saves in a private `mkdtemp` directory and commit them with `COPYFILE_EXCL`, closing the check/use race.
- Make MIME and direct IMAP fallback attachment writes use exclusive creation with owner-only mode `0600`, keeping both shipped save paths fail-closed.
- Release as 2.10.20 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 40 files, 575 tests passed
- `pnpm run lint` — 0 errors; 10 pre-existing `no-explicit-any` warnings on unchanged upstream lines
- `pnpm run format:check`
- `git diff --check`
- Corrected-head GitHub checks on `d8c813c53c444c745558371c63063926d519b776`: [CI run 31790903711](https://github.com/sweetrb/apple-mail-mcp/actions/runs/31790903711), [CodeQL check 94737560377](https://github.com/sweetrb/apple-mail-mcp/runs/94737560377), [version guard run 31790903687](https://github.com/sweetrb/apple-mail-mcp/actions/runs/31790903687), and [CodeQL analysis run 31790903697](https://github.com/sweetrb/apple-mail-mcp/actions/runs/31790903697) all passed; CodeQL reported no new alerts.
- Commit hook: `tsc --noEmit` and `vitest run` — 40 files, 575 tests passed
- Apple Mail integration tests: Not run; deterministic attachment-path coverage only.

Risk / Notes
- This changes shipped attachment-save behavior and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- Reusing a destination now fails instead of replacing the previous file; callers must choose a new path or remove the old file explicitly.
- Ready for maintainer review; no live Mail.app state was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.25",
+      "version": "2.10.26",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.25",
+      "version": "2.10.26",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.25",
+  "version": "2.10.26",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.10.25",
+  "version": "2.10.26",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.10.25",
+      "version": "2.10.26",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.25",
+  "version": "2.10.26",
   "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
   "author": {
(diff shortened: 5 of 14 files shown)


sweetrb:
The staging-directory-plus-`COPYFILE_EXCL` restructuring is the right shape — writing Mail.app's output into a private directory and committing it atomically genuinely closes the check/use window that the old `existsSync` preflight left open, and `mode: 0o600` on the MIME fallback is a real fix. Three things before merge.

**1. This PR opens a CodeQL alert and it's still open.**

`js/insecure-temporary-file` at `src/services/appleMailManager.ts:4170` — the MIME-fallback `writeFileSync(target.savedPath, attachment.data, { flag: "wx", mode: 0o600 })`. It fires because `fetch-attachment` routes through `saveAttachment` with an os-temp-derived directory, so the sink is reachable from `tmpdir()`. I think it's a false positive *given* `wx` + `0600` — that combination is exactly the prescribed mitigation — but it can't just be left sitting: either restructure so the fallback also writes into the `mkdtemp` staging directory and commits with `COPYFILE_EXCL` like the AppleScript path now does (which would make the two paths symmetric and is my preference), or dismiss the alert with that reasoning recorded.

**2. Refusing every existing destination is a behavior change that needs documenting.**

```js
if (existsSync(savedPath)) {
  if (lstatSync(savedPath).isSymbolicLink()) { throw … }
  throw new Error(`Refusing to overwrite existing file "${savedPath}"`);
}
```

Re-running the same `save-attachment` call now fails where it used to succeed — a plain retry after a partial failure, or saving the same attachment twice, both become errors. I'm fine with not-overwriting as the default, but it needs a CHANGELOG line that says so in those words, a README note on `save-attachment`, and ideally an explicit `overwrite: true` argument so the capability isn't simply gone. Right now the CHANGELOG entry doesn't mention it.

**3. The 0600 test asserts something the AppleScript path doesn't do.**

`creates MIME fallback files with owner-only permissions` covers the fallback, but the primary path's `copyFileSync` preserves the *staging* file's mode, which comes from Mail.app's own `save` — not necessarily 0600. So the permission guarantee the test implies holds on only one of the two routes. Either chmod after the copy or scope the claim.

Worth noting the staging directory is created inside `target.saveDirectory`, so a caller saving to a directory they can't write leaves a slightly different failure than before — the `mkdtempSync` failure path returning `false` with only a `console.error` is fine, just flagging it's a new failure mode.

jjoanna2-debug:
Addressed all three points. Both AppleScript and MIME paths now private-stage in the destination directory, commit with `COPYFILE_EXCL`, enforce final `0600`, and clean staging artifacts; this removes the insecure temporary-file path without suppressing the alert. README and CHANGELOG explicitly document fail-closed refusal of existing destinations; no overwrite escape hatch is added in this focused safety PR. Final head `b219cf0e778f978c6d87987041040495553a5ea7`; fresh CodeQL and CI are green. The alert-detail API remains 403 for this account, so I have not dismissed or waived anything.

sweetrb:
Merged and shipped in **2.10.26**. Thank you — the private-staging + `COPYFILE_EXCL` commit is a genuinely better shape than a check-then-write, and enforcing `0600` on the final file closes the window where an attachment lands world-readable.

One fix pushed to your branch: the two tests covering *this PR's own new guards* built their root with `mkdtempSync(os.tmpdir())`, which on macOS resolves to `/private/var/folders/…` — outside `ALLOWED_SAVE_ROOTS`. Both died on the pre-existing "outside allowed directories" check and never reached the destination-exists branch they were written for. Reverting the entire new guard left all six tests green.

Repointed them at an allowed root (as the other three tests in that file already do) and added an assertion that Mail.app is never handed the caller's destination path — only a staging path the exclusive copy then commits, which otherwise had no coverage either. Verified: dropping the overwrite refusal now fails 1, dropping the staging step fails 2.

PR #162, opened 2026-08-14
fix: cap IMAP attachment fetches

jjoanna2-debug:
Summary
- Cap IMAP attachment fetches at 25 MiB, matching the existing inline attachment limit.
- Reject a declared oversized BODYSTRUCTURE part before calling `download`.
- Enforce the same bound while streaming so a missing or inaccurate server size cannot cause an unbounded buffer; release as 2.10.21 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 40 files, 575 tests passed
- `pnpm run format:check`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 40 files, 575 tests passed
- Apple Mail integration tests: Not run; IMAP client tests use deterministic mock clients.

Risk / Notes
- This changes shipped IMAP `fetch-attachment` behavior and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- Attachments over 25 MiB now fail with an explicit limit error rather than being returned inline or written by `save-attachment`.
- Open for maintainer review; no live IMAP or Mail state was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.17",
+      "version": "2.10.21",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.17",
+      "version": "2.10.21",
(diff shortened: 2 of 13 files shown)


sweetrb:
Merged as 2.10.21. The part that made this an easy yes is that it bounds the stream *as well as* checking the declared `BODYSTRUCTURE` size — the declared size is server-supplied and can be absent or wrong, so the pre-download check alone would have been a guard that only works against honest servers. Checking both is what makes it an actual limit.

I added one Limitations-table line before merging, since the 25 MiB refusal is a new user-visible failure mode for `fetch-attachment`/`save-attachment` and belonged in the docs next to the count cap.

PR #163, opened 2026-08-14
fix: keep IMAP deletes recoverable

jjoanna2-debug:
Summary
- Make IMAP `delete-message` and `batch-delete-messages` recoverable-only.
- Refuse messages already in Trash instead of issuing `messageDelete`/EXPUNGE, and report the refusal through the existing single and batch error paths.
- Add regression coverage proving neither operation issues a destructive delete command; release as 2.10.22 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 40 files, 574 tests passed
- `pnpm run format:check`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 40 files, 574 tests passed
- Live IMAP integration: Not run; deterministic client mocks only.

Risk / Notes
- This changes shipped IMAP delete behavior and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- Deleting from Trash now fails safely; it cannot be used to empty Trash. There is no permanent-delete tool in this patch.
- Open for maintainer review; no live IMAP state was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.17",
+      "version": "2.10.22",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.17",
+      "version": "2.10.22",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.17",
+  "version": "2.10.22",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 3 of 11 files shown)


sweetrb:
Thanks for the careful read of the delete path — the recoverable-only principle is the right instinct for this server, and it's already the contract everywhere except the one case this PR removes.

**Declining this one, because that case is a deliberate, documented capability rather than an oversight.**

`trashUids`' expunge branch is the "empty from Trash" feature shipped in 2.8.5. From CHANGELOG.md:

> Delete now **moves the message to the account's Trash** … **Messages already in Trash are expunged (the "empty from Trash" case).**

So the delete contract isn't recoverable-only — it's *recoverable everywhere except when the caller has explicitly addressed a message that is already in the Trash*, which is the one place where "delete" can only reasonably mean "remove". This PR replaces that with a hard error and leaves no way to empty Trash through the server at all.

Two specifics:

1. **The safety gain is narrow.** Reaching the expunge branch requires the caller to pass an `imap:` id whose mailbox component *is* the resolved Trash path. That id came from listing Trash. It isn't a path a confused deputy stumbles onto while trying to delete an inbox message — for that case the PR changes nothing, because `srcPath != dest` and the move branch already runs.
2. **The regression is real.** Autonomous hygiene flows that empty Trash on a TTL lose their only route, and get a hard failure rather than a no-op.

If you want to keep a guard here, the shape I'd take is an explicit opt-in — a `permanent: true` argument on `delete-message`, or an `APPLE_MAIL_MCP_ALLOW_EXPUNGE` env gate — defaulting to refusing, so the default posture matches your PR while the documented capability survives. Happy to look at that if you want to open it.

The test restructuring around `imapBatchDelete` is good regardless, and the diagnosis that the two branches deserve distinct error text stands.

jjoanna2-debug:
Thanks, agreed. I am leaving #163 closed because empty-Trash expunge is deliberate and documented; there is no code change on this PR. I will treat an explicit `permanent: true` or environment-gated purge design as a separate follow-up rather than changing delete semantics here.

PR #164, opened 2026-08-14
fix: disable new mail rules by default

jjoanna2-debug:
Summary
- Make new `create-rule` entries disabled by default so an automation that can delete or move future mail has a review point.
- Preserve explicit opt-in with `enabled: true`; update the tool schema, type documentation, and manager behavior.
- Add deterministic AppleScript-shape tests and release as 2.10.23 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 41 files, 575 tests passed
- `pnpm run format:check`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 41 files, 575 tests passed
- Live Mail integration: Not run; deterministic AppleScript mocks only.

Risk / Notes
- This changes shipped `create-rule` behavior and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- Existing rules are unchanged; only newly created rules default to disabled. Callers that require immediate automation must pass `enabled: true` explicitly.
- Open for maintainer review; no live Mail state was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.26",
+      "version": "2.10.27",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.26",
+      "version": "2.10.27",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.26",
+  "version": "2.10.27",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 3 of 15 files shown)


sweetrb:
The reasoning holds up: `create-rule` is the one tool here whose output keeps acting after the call returns, and a rule with `delete: true` that starts live is a genuinely different risk from every other write in this server. `enable-rule` already exists, so creating disabled doesn't strand anything — the review step is one extra call, not a trip into Mail's settings. I checked that before forming a view, since it would have been the deciding objection.

Two things needed before merge.

**1. The README still documents the old default, and this PR doesn't touch it.**

```
| `enabled` | boolean | No | Whether the rule is enabled on creation (default `true`) |
```

That line is now wrong. It needs to flip to `false` and say what to do next — create, review with `list-rules`, then `enable-rule`. The repo has a docs guard that checks documented behavior against reality, so this is also the kind of drift that's meant to be caught rather than shipped.

**2. The CHANGELOG entry needs to name it as breaking.**

Anyone whose flow is "create-rule, done" gets a rule that silently does nothing after this. That's the right trade, but it has to be findable by someone debugging exactly that surprise — an entry under `### Changed` saying the default flipped and pointing at `enable-rule`.

`enabled: opts.enabled === true` is the correct implementation of the flip, and the `types.ts` doc comment was updated, so the code half is fine.

Flagging for @sweetrb rather than merging on my own: this is a default-flip on a published package, so it's your call whether the safety is worth the surprise. My read is that it is, given `enable-rule` makes the recovery a one-liner.

jjoanna2-debug:
Addressed both docs points. README now says `enabled` defaults to false and directs callers through create, review with `list-rules` or Mail.app, then explicit `enable-rule`; CHANGELOG marks the default flip as a breaking behavior change. Runtime, tests, and build are aligned. Final head `387d6034af5513230fdb4e9269aee2b739d1051a`; fresh CI and CodeQL are green.

sweetrb:
Merged and shipped in **2.10.27**. Thank you — this one went in unchanged.

I reviewed it adversarially on two points and both came back clean: the tool description does convey the new default (the field-level `.describe()` renders into the advertised schema, so a caller sees it), and adding `enabled` to the response would *not* have tripped the `additionalProperties: false` trap the way I first suspected. Neither concern survived contact with the code.

Defaulting a newly created mail rule to off is the right default for an agent-driven tool — a rule is a standing instruction that keeps acting long after the call returns, so "create it, let the human look, then enable" is the correct order. `types.ts`, the zod schema and the README all agree on the new default, which is what made this cheap to verify.

PR #165, opened 2026-08-14
fix: require encrypted mail transport

jjoanna2-debug:
Summary
- Require STARTTLS for SMTP when implicit TLS is not selected (`requireTLS: true`).
- Require STARTTLS for non-implicit IMAP connections (`doSTARTTLS: true`) instead of accepting an unencrypted downgrade.
- Add transport-option tests and release as 2.10.24 with the rebuilt bundle and synchronized plugin manifests.

Verification
- `pnpm run build`
- `pnpm test` — 40 files, 574 tests passed
- `pnpm run format:check`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 40 files, 574 tests passed
- Live SMTP/IMAP integration: GitHub IMAP integration passed in [run 31779336104](https://github.com/sweetrb/apple-mail-mcp/actions/runs/31779336104); no local provider credentials were used, so injected transport and pure option tests remain the local reproducible checks.

Risk / Notes
- This changes shipped SMTP/IMAP connection behavior and updates `CHANGELOG.md`, `package.json`, committed `build/index.js`, and plugin manifests.
- Servers that do not support STARTTLS now fail rather than sending credentials or mail over plaintext; implicit-TLS configurations are unchanged.
- No local live mail transport was contacted; the authenticated GitHub IMAP integration check passed.
- Open for maintainer review; this PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.24",
+      "version": "2.10.28",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.24",
+      "version": "2.10.28",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.24",
+  "version": "2.10.28",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.10.24",
+  "version": "2.10.28",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.10.24",
+      "version": "2.10.28",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.24",
+  "version": "2.10.28",
   "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
   "author": {
     "name": "Rob Sweet",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,10 +133,10 @@ jobs:
       greenmail:
         image: greenmail/standalone:2.1.0
         ports:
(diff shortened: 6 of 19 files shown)


sweetrb:
This is the strongest of the batch and I want it in. I checked the load-bearing claim rather than taking it on faith: `doSTARTTLS` is a real ImapFlow option in the pinned 1.4.1 (`lib/imap-flow.d.ts:107`, honored at `lib/imap-flow.js:1494`), and opportunistic-by-default really is the prior behavior — so `doSTARTTLS: !cfg.secure` closes an actual downgrade path rather than being decorative. `requireTLS: !cfg.secure` on the Nodemailer side is the correct counterpart. Two blockers, neither about the security logic.

**1. No escape hatch for plaintext IMAP, and the failure will be opaque.**

Anyone pointing this at a LAN dovecot/Courier without TLS, or at a stunnel-fronted server, goes from working to a connection error with no hint that a policy decision caused it. Please add an explicit opt-out — `APPLE_MAIL_MCP_IMAP_ALLOW_PLAINTEXT=1` / the SMTP equivalent — that leaves the secure default exactly as you've written it, and make the thrown error name the variable. Refusing by default is right; refusing with no documented way forward is what generates "it just stopped working" reports.

Related: `buildImapConnectionOptions` sets `doSTARTTLS: false` when `secure` is true, which is required (ImapFlow throws on `secure && doSTARTTLS`, `imap-flow.js:1513`) — worth a one-line comment saying so, because it reads like a no-op and someone will "simplify" it later.

**2. It can't merge from a fork as-is, and that's on our side, not yours.**

It touches `.github/workflows/ci.yml`, which this repo can't take through a fork PR. I'll carry it forward as a continuation PR on an in-repo branch with your commits cherry-picked and a `Co-authored-by` trailer, once the escape hatch above is in — either push it here and I'll pick up the final version, or say the word and I'll add it myself.

One note on the integration fixture: moving GreenMail to IMAPS means `rejectUnauthorized: false` for its self-signed cert, so `imap-integration` no longer exercises certificate verification at all. That's an acceptable trade for testing the STARTTLS/implicit-TLS split, but the comment should say the fixture deliberately drops cert verification, so nobody later reads a green `imap-integration` as evidence that verification works.

Docs are also missing — the TLS requirement and the new variable belong in the IMAP/SMTP setup section, not only in the CHANGELOG.

jjoanna2-debug:
Addressed both blockers and the fixture/docs notes. Added `APPLE_MAIL_MCP_IMAP_ALLOW_PLAINTEXT=1` and the SMTP equivalent as explicit unsafe opt-outs; secure defaults remain fail-closed, errors name the variable, and the `secure=true` with no-STARTTLS case is commented. IMAP IDLE now uses the same policy helper. README and CHANGELOG cover TLS and opt-outs; GreenMail comments scope `rejectUnauthorized: false` to the self-signed fixture. `.github/workflows/ci.yml` remains a maintainer-continuation or cherry-pick surface as you described. Final head `63411740a350eb33d36a11f28b3a5f6b41136401`; fresh CI, CodeQL, and IMAP checks are green.

sweetrb:
Landed as #171 — merged, shipping in **2.10.28**. Thank you for this one; requiring encrypted transport by default was the right call, and naming the escape hatches in the failure messages is what makes it survivable for people on odd relays.

**Why a new PR rather than merging this one.** It touches `.github/workflows/ci.yml`, and fork PRs can't modify workflows in this repo — nothing to do with the change itself. Your four commits were cherry-picked with authorship intact and the squash carries a `Co-authored-by` trailer, so the credit is on the merge commit.

**One fix added on top,** because it inverted the intent of your own escape hatch:

```ts
doSTARTTLS: !cfg.secure && !cfg.allowPlaintext   // -> literal `false` when the hatch is on
```

ImapFlow reads that field as a tri-state, and `false` does not mean "don't require it" — it means **never STARTTLS, even if the server advertises it** (`imap-flow.js:1521`, and the JSDoc at `:146`). Before this PR the option was absent, i.e. opportunistic. So the opt-out didn't restore the prior behaviour, it went below it.

The sharp edge is that following your own remediation makes things worse: two accounts on port 143, both offering STARTTLS; account B's cert expires; the new error tells the user to set `APPLE_MAIL_MCP_IMAP_ALLOW_PLAINTEXT=1`; account A — which had been upgrading fine — now sends its Keychain password in cleartext on every pooled connect.

Changed to `cfg.secure || cfg.allowPlaintext ? undefined : true`, so the hatch can only ever relax a *requirement*. `imapIdle.ts` shares `buildImapConnectionOptions`, so it came along. Three of your assertions expected `doSTARTTLS: false` — they encoded the defect, so they were updated rather than deleted, and reverting the one-line change fails 4 tests.

Your SMTP half was checked and is genuinely correct: nodemailer's `requireTLS: false` still upgrades opportunistically, since only `ignoreTLS` suppresses it. The asymmetry was in imapflow's semantics, not in your reasoning.

PR #166, opened 2026-08-14
chore: pin Codex MCP runtime

jjoanna2-debug:
Summary
- Pin the Codex plugin MCP launcher to the exact current published package `apple-mail-mcp@2.10.17` instead of resolving the moving `latest` tag on every launch.
- Document that the pin must be updated deliberately during an upgrade.

Verification
- JSON parse of `codex/.mcp.json`
- `git diff --check`
- Commit hook: `tsc --noEmit` and `vitest run` — 40 files, 573 tests passed
- npm registry readback: `latest` is `2.10.17`

Risk / Notes
- This is a docs/config-only governance change; no package version bump or rebuilt bundle is required under `CONTRIBUTING.md`.
- The pin reduces unreviewed runtime drift but requires a deliberate update when maintainers release a new version. Other hosts remain unchanged.
- Open for maintainer review; no mail state or external automation was changed. The PR is not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.22",
+      "version": "2.10.23",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.22",
+      "version": "2.10.23",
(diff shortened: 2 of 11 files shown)


sweetrb:
Merged as 2.10.23. Right diagnosis: a bare `npx -y apple-mail-mcp` meant the plugin manifest's version and the runtime it actually launched were two independent claims, and only one of them was auditable.

I added the piece that keeps it true — `scripts/sync-plugin-version.mjs` now rewrites the pin from `package.json` on every bump. Without that the pin rots at the next release and is *worse* than no pin, because it asserts a precision it no longer has. CI already runs the script's `--check` mode, so drift fails the PR; I verified the guard by mutating the pin and confirming it exits 1 naming `codex/.mcp.json` rather than assuming it would.

PR #168, opened 2026-08-14
fix: constrain outbound attachment reads

jjoanna2-debug:
## Summary

- Constrain outbound string attachment reads to canonical regular files inside the default Desktop, Documents, Downloads, or temporary roots, closing the arbitrary-local-file disclosure path identified by the security audit.
- Apply the same resolver to both AppleScript and SMTP attachment paths, with an explicit `APPLE_MAIL_MCP_ATTACHMENT_READ_ROOTS` absolute-root allowlist for deliberate expansion.
- Replace the stale security-test copy of attachment validation with tests against the real production boundary, document the behavior, and publish patch version 2.10.25 with the rebuilt bundle.

## Verification

- `pnpm test` — 41 test files, 570 tests passed.
- `pnpm run typecheck` — passed.
- `pnpm run lint` — 0 errors, 10 existing warnings.
- `pnpm run format:check` — passed.
- `pnpm run build` — passed; committed `build/` refreshed.
- `git diff --check` — passed.
- Commit and push hooks — typecheck and 570 tests passed.
- Live Mail.app integration — Not run; this environment does not have a configured live Mail.app integration fixture.

## Risk / Notes

- Public-facing files changed: `README.md`, `CHANGELOG.md`, `package.json`, plugin manifests, and committed `build/`.
- This is intentionally compatibility-restricting: existing attachment paths outside the default roots must be authorized with `APPLE_MAIL_MCP_ATTACHMENT_READ_ROOTS`, a colon-separated list of absolute roots. Inline base64 attachments are unchanged.
- Symlinks are canonicalized before the root check, so a link escaping an allowed root is rejected. A residual same-host TOCTOU window remains between validation and the downstream reader opening the canonical path.
- No live GitHub settings or automations were changed. Open for maintainer review; not merged, shipped, accepted, or maintainer-approved.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.28",
+      "version": "2.10.29",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.28",
+      "version": "2.10.29",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.28",
+  "version": "2.10.29",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.10.28",
+  "version": "2.10.29",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.10.28",
+      "version": "2.10.29",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.10.28",
+  "version": "2.10.29",
   "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
   "author": {
     "name": "Rob Sweet",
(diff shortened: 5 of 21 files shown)


sweetrb:
The threat you're describing is real and I want a fix for it: `send-email` taking an arbitrary absolute path does make this server an exfiltration primitive for a model that has been talked into it, and the asymmetry with `save-attachment` (whose *write* root has been constrained for a while) is a fair catch. `resolveAttachmentReadPath` is also well built — canonicalizing with `realpathSync` *before* the root test is the part people usually get wrong, and you got it right.

**The default root set is what blocks this.** Desktop / Documents / Downloads / tmp excludes essentially every working directory a developer actually attaches from — source trees, `/Volumes/*` external and network volumes, and iCloud Drive (`~/Library/Mobile Documents/…`), which is where a lot of macOS users' documents genuinely live. Shipping that as the default turns a large fraction of ordinary `send-email` calls into hard errors, and the escape hatch is an env var that has to be set on the MCP host config and restarted — not something a user discovers from the error at the moment it bites.

There's also a soundness wrinkle: `tmpdir()` and `/tmp` are in the default allowlist and are world-writable, so any local process can stage a file into an allowed root. That doesn't make the control useless, but it does mean the boundary is not "user content only", and the doc comment shouldn't imply it is.

Two directions I'd accept, either is fine:

- **Denylist, not allowlist** — default to allowing `homedir()` and `/Volumes`, and refuse the sensitive subtrees explicitly (`~/.ssh`, `~/Library/Keychains`, `~/Library/Application Support/*/config.json`, `~/.aws`, `~/.config/gh`, dotfiles generally). That kills the actual exfiltration targets while leaving normal attachment flows working.
- **Default open, opt-in restriction** — keep today's absolute-and-exists check as the default and let `APPLE_MAIL_MCP_ATTACHMENT_READ_ROOTS` *impose* an allowlist when set. Security-conscious deployments get the boundary; nobody's existing setup breaks.

I lean toward the first: it's on by default, which is the point, and it's the version where the "allowed roots" error message will almost never be seen by someone doing something legitimate.

Whichever way it goes, this needs a README section (not just the schema `.describe()`) and a CHANGELOG entry calling it a breaking change for attachment paths, since it is one.

Holding this one for @sweetrb to weigh in on the default posture before more work goes into it — the code is not the hard part here, the default is.

jjoanna2-debug:
Addressed the default-root concern with the denylist design. Defaults now allow `homedir()`, `/Volumes`, and temporary roots, but reject hidden files and known credential or configuration paths; extra absolute roots are additive and cannot bypass protected-path checks, and symlink escapes are canonicalized and rejected. README calls out `/tmp` as a world-writable convenience root, plus the breaking path behavior and environment setting; tests cover ordinary files, denylist, explicit roots, and symlink escape. Final head `67ad2113499335ef91fb9678abff497684c27b56`; fresh CI and CodeQL are green.

sweetrb:
Merged and shipped in **2.10.29**. Thank you — this was the most valuable of the batch. `buildAttachments` accepting any absolute path that exists really did make `send-email` an arbitrary local-file read primitive for anything driving this server, and the denylist-over-allowlist redesign after the first review round was the right correction: confining reads to `Documents/Desktop/Downloads` would have broken ordinary use for a threat that a hidden-file + credential-location denial covers better.

One fix pushed to your branch: the policy canonicalized with `fs.realpathSync`, which resolves symlinks but **preserves the caller's casing**. Only `fs.realpathSync.native` returns the true on-disk name, and macOS APFS is case-insensitive by default — so a denied location was reachable by respelling one segment:

```
~/Library/Keychains/login.keychain-db   -> DENY
~/library/keychains/login.keychain-db   -> ALLOW   (same 4.1 MB file)
```

I ran your policy code against both spellings on a real machine and got exactly that. `Keychains` is the only deny root not also caught by the hidden-segment rule, which is why it was the reachable one.

Now canonicalized through one `realpathSync.native` helper — candidate path, configured roots, deny roots and the home prefix — so exact comparison stays correct on case-sensitive volumes too, where the respelling simply doesn't exist. Regression tests assert the policy reports the on-disk spelling and denies every spelling the filesystem accepts; reverting to the non-native call fails 2, including the real keychain file being let through.

PR #169, opened 2026-08-14
test: share validation schemas

jjoanna2-debug:
## Summary

- Move the production message-id, batch-id, date-filter, and attachment schemas into a shared module so the MCP tool registrations and security tests use the same validation contract.
- Remove the stale security-test schema copies and add coverage for injection-safe IMAP ids and the real attachment union/cap.
- Publish patch version 2.10.26 with the rebuilt bundle; runtime validation behavior is preserved.

## Verification

- `pnpm test` — 40 test files, 575 tests passed.
- `pnpm run typecheck` — passed.
- `pnpm run lint` — 0 errors, 10 existing warnings.
- `pnpm run format:check` — passed.
- `pnpm run build` — passed; committed `build/` refreshed.
- `git diff --check` — passed.
- Commit and push hooks — typecheck and 575 tests passed.
- Live Mail.app integration — Not run; this is a schema/test-contract change and the local environment has no live Mail.app fixture.

## Risk / Notes

- Public-facing files changed: `CHANGELOG.md`, `package.json`, plugin manifests, and committed `build/`; `README.md` is unchanged.
- This is a source-organization change intended to prevent security tests from silently validating a different contract than the server advertises. The schema expressions and messages are preserved.
- No live GitHub settings or automations were changed. Open for maintainer review; not merged, shipped, accepted, or maintainer-approved.
- This branch is intentionally based on current upstream main; it may need a normal rebase if another version-bumping hardening PR lands first.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.21",
+      "version": "2.10.22",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.10.21",
+      "version": "2.10.22",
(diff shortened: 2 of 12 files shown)


sweetrb:
Merged as 2.10.22. This one was worth more than the "test:" prefix suggests — `security.test.ts` had been re-declaring `MESSAGE_ID_SCHEMA` as `/^\\d+$/`, so the suite was asserting against a schema that rejects the `imap:…` form the server has accepted for a long time. The tests were green and testing fiction. Collapsing them onto one imported module means that particular drift can't recur.

Issue #176, opened 2026-08-15
Collateral snapshot goes unavailable on large mailboxes — chunk it instead of giving up


sweetrb:
Splitting this out of #155, where it was twice described to @scottstern0325 as "tracked separately" — it was not actually tracked anywhere. Filing it so that statement becomes true.

## The problem

The collateral diff added in #157 snapshots the source mailbox before and after a destructive operation and names every message that disappeared, **including ones that were never in the id list**. It is the only mechanism that can attribute an unrequested departure, which is the entire symptom #155 exists for.

On a 29-id batch, @scottstern0325 got:

```
snapshot: "unavailable"
skipReason: "Mail would not produce the after snapshot for this mailbox"
```

— and during that same batch an Amazon shipment notice they had not named left INBOX and landed in Trash. Because the snapshot was unavailable, that departure is **un-attributable**: this server moving it, a Mail rule firing, a server-side filter, or another device all read identically. It was fully recoverable from Trash, so no data was lost, but the evidence was.

## Why this is worse than an ordinary flaky path

Taking the snapshot is the most expensive thing the operation does, and the cost scales with mailbox size. So "Mail wouldn't answer" gets **more likely as the batch and mailbox get bigger** — the one mechanism meant to catch collateral damage is least reliable exactly when the blast radius is largest, and when a human is least able to eyeball the result by hand. That correlation is the defect, not the individual failure.

## What to do

Chunk the snapshot rather than issuing one whole-mailbox request and giving up when Mail declines it. Sketch, not a spec:

- page the mailbox in bounded slices instead of materializing the whole message list in one AppleScript round trip;
- on a slice that fails, retry that slice rather than abandoning the whole snapshot;
- if some slices still fail, report a **partial** snapshot — the set of messages that were successfully observed, with the unobserved range named — instead of the current all-or-nothing `unavailable`. A partial diff that says "nothing unrequested left the 400 messages I could see, and I could not see the other 120" is strictly more useful than no diff, and honest about the gap;
- keep `status: unknown` semantics for whatever is genuinely unobserved, so it never claims an operation was clean.

## Guard requirement

Per the project's regression rule, the guard for this must be shown failing before the fix: construct or simulate a mailbox large enough that the current implementation returns `snapshot: "unavailable"`, confirm the test fails against today's code, then fix. A guard that has never seen the bug it targets does not count.

## Related

- #155 — the unexplained `over` residual this instrumentation exists to catch (still open, still unreproduced).
[...]

PR #186, opened 2026-08-16
fix(audit): a stale-low count no longer fabricates collateral findings (#179)

### LINKED ISSUE #179: Collateral snapshot is count-bounded, so a stale-low count truncates it silently and can fabricate `disappeared`
Split out of #155 / #178 so it is tracked rather than living in a docstring.

## The defect

`snapshotFragment` bounds its enumeration by the mailbox count it was handed:

```applescript
repeat while _sLo <= _ca
  set _sHi to _sLo + _sChunk - 1
  if _sHi > _ca then set _sHi to _ca
```

(`src/services/appleMailManager.ts`, and the same for `_cb` on the before phase.)

Positions past that bound are **never requested**. Only a *failed* slice enters `_sMiss`, so an unrequested tail leaves no trace: `_sStatus` stays `"ok"`, `holes.length === 0` in the diff, and the record claims a complete observation.

#155 established that Mail's count can lag the mutation. If it ever reads **low**, the after-enumeration is truncated, and every message past the bound is present in `before`, absent from `after` — landing in `disappeared`, and if the caller never named it, in `unrequested`, which the audit log documents as *"IS the #155 symptom, with names attached."*

That is the fabricated finding the #176 contract promises is impossible:

> a fabricated finding, which is worse than the gap it papers over

## Status

- **Not observed in the field.** Every #155 reading is stale-**high**. This is reachable in principle, not evidenced.
- **Cannot currently fabricate under stale-high**, which is the direction we have data for: a high bound over-requests, and the out-of-range slice either fails (→ recorded as a hole, correct) or returns nothing.
- Documented as a caveat in README and on `CollateralDiff` in 2.11.0 — `"snapshot": "ok"` is stated to be necessary but not sufficient.

## Why it wasn't fixed in #178

The obvious fix is to probe one slice past the bound and treat a non-empty result as evidence the count was low. Whether `messages i thru j of mb` **clamps or raises** when `j` exceeds the true length is plausibly backend-dependent (iCloud/IMAP vs Exchange vs POP vs On-My-Mac). Getting it wrong the other way is a strict regression for audit-log users: an over-request that raises would collapse otherwise-fine snapshots to `partial`.

Deciding that from a single account would repeat the mistake #155 just corrected. Needs the semantics established across account types first, or a runtime detection that does not assume.

## Guard requirement

Per the project rule, the guard has to be shown failing first. The simulator can already model this — `staleAfterCountBy` (added in #178) makes the count and the listing disagree — but note the simulator currently slices by the fake mailbox's array length rather than by the count the generated script emitted, so **it has to be changed to derive its slice range from the emitted count**, exactly as `snapshotFragment` does. Until that changes, a test for this bug would pass vacuously against broken code.

Related: #155, #176, #178.

sweetrb:
**Closes #179.** Combines the two planned steps — the simulator fidelity work is the prerequisite for the guard, and merging it alone would add no test, so they ship together.

## The defect

`snapshotFragment` bounds its enumeration by the count Mail just reported:

```applescript
repeat while _sLo <= _ca
```

#155 established that the count can lag the mailbox, and the count and the snapshot are **consecutive Apple Events in one script with nothing between them** — co-stale, not independent instruments.

When the count reads **low**, positions past the bound are **never requested**. Only a *failed* slice enters `_sMiss`, so an unrequested tail leaves no trace: `_sStatus` stays `ok`, `holes.length === 0`, and the record claims a complete observation. Those messages are then present in `before`, absent from `after`, and land in `disappeared` — and if the caller never named them, in `unrequested`, which the audit log documents as *"IS the #155 symptom, with names attached"*.

That is exactly the fabricated finding the #176 contract promises is impossible: an innocent message put in front of someone mid-incident as evidence of data loss.

Regression from #177 (2.10.33) — v2.10.32 read `id of messages of mb` unbounded.

## The fix

Probe exactly **one** position past the bound. One rather than a slice, because an out-of-range *range* raises as a whole, so an over-requested slice cannot distinguish "nothing there" from "the count was low by more than a chunk".

A message found there means the count was low; the unread tail is recorded as a hole, which makes the snapshot `partial` under the **existing** rules and withholds exactly the halves a truncation would poison. **No new status, no new gating** — it reuses the #176 contract rather than inventing semantics beside it.

The probe cannot fire spuriously:
- with an accurate count the probed position does not exist and the `try` swallows the raise;
- a specifier that **clamps** instead of raising hands back an id this enumeration already recorded, which is ignored.

Whether Mail's `messages i thru j` clamps or raises is *still* not established across backends (iCloud/IMAP vs Exchange vs POP vs On-My-Mac), so both are handled rather than assumed — which was the reason #178 declined to fix this.

## The simulator had to be fixed first

`emitSnapshot` sliced by the fake mailbox's **real length** rather than by the count the generated script emitted, so the bug was **not expressible in a test at all** — a guard for it would have passed vacuously against broken code. That is the same trap the retracted `falseOkWarning` fell into, and #179 called it out explicitly as a prerequisite.

It now derives its range from the emitted count and models an out-of-range slice **raising**, with an `outOfRangeSliceClamps` knob for the other semantics. `staleAfterCountBy` consequently bounds the after-*snapshot* as well as the after-*count*, which is the point.

## Verification

Guard checked in **both** directions:

| | result |
|---|---|
| unfixed script + new simulator | **RED** — `unrequested` = `[75814, 75815]`, two messages that never moved, with `snapshot: "ok"` |
| fixed script | **GREEN** |

Exactly one test flips; the other 48 forensics tests are unchanged in both directions.

Generated AppleScript validated with `osacompile` (exit 0) — the probe renders correctly in both phases (`_cb` before, `_ca` after).

Suite: 665 passed / 46 files. `typecheck` clean, `lint` 0 errors, `format:check` clean.

Closes #179

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.13.0",
+      "version": "2.13.1",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.13.0",
+      "version": "2.13.1",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.13.0",
+  "version": "2.13.1",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.13.0",
+  "version": "2.13.1",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.13.0",
+      "version": "2.13.1",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.13.0",
+  "version": "2.13.1",
   "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
   "author": {
     "name": "Rob Sweet",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,52 @@
 ## [Unreleased]
 
+## [2.13.1] - 2026-08-16
+
+Closes #179 — a regression introduced by the snapshot chunking in #177 (2.10.33).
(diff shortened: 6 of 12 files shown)


sweetrb:
Filed the stale-HIGH counterpart as #187 — same root cause (the loop is bounded by a count that can lag), opposite direction, different failure mode, and deliberately **not** folded into this PR.

Worth noting because it inverts the severity of #176: a stale-HIGH count makes the single slice of a **sub-chunk** mailbox raise as a whole, so `_sPairs` ends up empty and the snapshot collapses to `unavailable` with no holes and no warning. Verified with the simulator this PR adds: `{"snapshot":"unavailable","disappeared":null,"unrequested":null,"warnings":[]}`.

That direction is the one #155 actually evidences, so the instrument is off precisely where it is needed. Its fix is a clamping retry, which is a different change from the probe here.

PR #198, opened 2026-08-21
fix(applescript): distinguish nested mailboxes with duplicate leaf names

### LINKED ISSUE #197: AppleScript mailbox lookup conflates nested mailboxes with the same leaf name
## Description

On my Exchange account, `mailboxes of account` includes both a top-level `Inbox` and a nested `Archive/Inbox`. Apple Mail reports `name of mailbox` as `Inbox` for both.

The AppleScript backend currently caches and resolves that leaf name, so the two mailboxes are indistinguishable and an operation scoped to `Inbox` can target the archived mailbox instead.

This differs from the assumption in #14 that nested mailboxes are named by path. The ambiguity handling from #137 applies to IMAP, not the AppleScript backend.

## Expected behavior

`Inbox` should resolve to the top-level mailbox. The nested mailbox should be addressable as `Archive/Inbox`, and an ambiguous leaf name should be rejected instead of choosing one.

## Environment

- macOS Mail
- Exchange account

## Fix
I have a fix #198 that reconstructs each mailbox's account-relative path from its container chain, prefers exact paths, and rejects ambiguous leaf names. It includes regression tests and a live check against the affected account.

Tal-Golan:
## Description

I ran into this on an Exchange account with both a top-level `Inbox` and a nested `Archive/Inbox`. Mail returns both mailbox names as `Inbox`, so the AppleScript backend can resolve the wrong one.

I have a fix that reconstructs each mailbox's account-relative path from its container chain. Exact paths win, unique leaf names still work, and ambiguous leaf names return an error with the matching paths. The existing response fields are unchanged.

Fixes #197.

## Testing

- [x] 704 unit tests pass
- [x] Lint, typecheck, formatting, and build pass
- [x] Checked against the affected Exchange account


sweetrb (changes requested):
Thanks for this — the bug in #197 is real and well diagnosed, and reconstructing the account-relative path from the container chain is the right approach. The extra detail that `mailboxes of account` is recursive while `name of` is only the leaf is correct and worth having written down.

I can't merge this as-is, though. The path walk hard-fails against a real Mail.app, and because every test in this repo mocks `executeAppleScript`, the 704 green tests can't see it. Details and a verified fix below.

---

## Blocker 1 — the walk never terminates, then raises `-1728`

```applescript
if _parentClass is account or _parentClass is application then exit repeat
```

Mail.app never reports a bare `account` class. It reports the concrete subclass — `imap account`, `pop account`, `exchange account`, `mac account`. Probed live:

```
hop 1: name=[Gmail]                   class=container     is account=false
hop 2: name=robert.b.sweet@gmail.com  class=imap account  is account=false
  -> container of this raised: (-1728) Mail got an error:
     Can't get container of account id "72F83AFA-1F45-4FB8-8772-7532EB0148D1".
```

So the guard is dead code. The loop prepends the **account's own name** to the path, then calls `container of <account>`, which raises. Same `list-mailboxes` call, same machine, two builds:

```
main    (2.16.1)  Found 14 mailbox(es): Important, All Mail, INBOX, Boat, ...
#198    (2.16.2)  Could not list mailboxes for "robert.b.sweet@gmail.com":
                  Mail got an error: Can't get container of account id "72F83AFA-...".
```

This is not Exchange-specific — it hits every account type. `list-mailboxes` returns empty, `fetchMailboxNames` caches `[]` (so the new ambiguity guard is itself inert), and `move-message` / `batch-move-messages` / `get-message` fail or return null.

I suspect your Exchange check passed because the `try`-wrapped inlinings swallow the raise — but they swallow it *after* the account name has already been prepended, so those paths silently emit `work@example.com/Archive/Inbox` into `Message.mailbox`.

## Blocker 2 — "On My Mac" raises one step earlier

For local-store mailboxes, `container of` returns a phantom `«class mact»` account reference rather than `missing value`, and `class of` **that** raises `-1728` before any comparison runs. Guarding only the `is account` comparison doesn't save this path.

## Verified fix

Whitelist the classes you're willing to walk instead of blacklisting terminators, and guard each accessor. Note nested parents report `class` = `container`, not `mailbox`, so both terms are required:

```applescript
set OUT to name of MB
set _pathParent to missing value
try
  set _pathParent to container of MB
end try
repeat while _pathParent is not missing value
  set _parentClass to missing value
  try
    set _parentClass to class of _pathParent
  end try
  if _parentClass is not mailbox and _parentClass is not container then exit repeat
  set OUT to (name of _pathParent) & "/" & OUT
  set _pathNext to missing value
  try
    set _pathNext to container of _pathParent
  end try
  set _pathParent to _pathNext
end repeat
```

I patched your branch with this and rebuilt. Both cases now work:

```
Gmail account : [Gmail]/All Mail, [Gmail]/Trash, INBOX, Boat, Financial, ...
On My Mac     : Import, SendLater, Outbox, Deleted Messages
```

A nice side effect: the AppleScript backend's paths now agree exactly with what the IMAP backend already returns.

Please also give every inlining a `try` with fallback to the leaf name, so a future Mail change degrades to today's behaviour instead of emptying `list-mailboxes` and the resolver cache.

---

## Blocker 3 — the new throw can fire *after* a destructive op has run

`resolveAppleMailboxPath` throws on an ambiguous leaf, and `resolveMailbox` is called on the **post-operation** forensics path (`recordSingleForensics`, and the `sameMailbox` closures).

Account with `Personal/Receipts` and `Work/Receipts`, no top-level `Receipts`: `move-message {id: 42, mailbox: "Archive"}` for an unlocated message succeeds — Mail moves it — then forensics calls `resolveMailbox("Receipts", ...)` on the leaf from the RECON record, which throws. The tool reports `Error moving message: Mailbox "Receipts" is ambiguous ...` while the message is sitting safely in `Archive`. With `batch-move-messages` that misreports a whole batch.

Fix: never call the throwing resolver on a post-op path. Either make `reconEmitFromMessage` use `mailboxPathFragment` so RECON records carry canonical paths too, or split the resolver into a throwing `resolveOrThrow` for pre-op input validation and a non-throwing variant for forensics.

## Blocker 4 — `rename-mailbox` can orphan a mailbox

`createMailbox` runs before `resolveMailbox(oldName)`. With `Personal/Receipts` + `Work/Receipts`, `rename-mailbox {oldName: "Receipts", newName: "Invoices"}` creates `Invoices`, *then* throws — leaving an empty orphan and no rollback. Resolve both names before creating anything, and return `{success: false, error}` rather than throwing.

## Blocker 5 — `rowsIncludeMailbox` has no test coverage

This is the riskiest change in the PR: field 7 of a row is either a mailbox path or an attachment boolean, decided solely by the new flag. Two independent mutations of `parseMessageList` survive the entire suite. One of them makes `rememberLocation` cache the literal string `"true"` as a message's mailbox, so every later by-id read/reply/delete targets `mailbox "true"`.

Related, and the reason this needs a test rather than a re-read: `let rowsIncludeMailbox = mailbox === undefined` desynchronizes from the branch that actually emits the rows, which is gated on `if (mailbox)`. Pass `mailbox: ""` (schema-legal, nothing rejects it) and you get 8-field rows parsed by neither branch — every message reports `mailbox: "INBOX"` and `hasAttachments: false`, and `rememberLocation` caches the wrong mailbox. Since Mail.app numeric ids are unique only *within* a mailbox, a follow-up delete/move on one of those ids can bind a different message. `!mailbox` at both sites fixes it.

Please add a `parseRows` test covering all four shapes (7/8 fields × flag true/false), plus an Exchange-fixture test that asserts on the **generated script** for the actual #197 scenario — right now `resolveMailbox` can be reduced to the identity function and the suite stays green.

---

## Smaller things

- `reconEmitFromMessage` still emits the leaf, so `countDelta` and the audit log can report two entries for one physical mailbox (`Archive/Inbox` and `Inbox`), one of which no message was ever in.
- The ambiguous-id diagnostics still print leaf names, which makes the error for the exact `Inbox` / `Archive/Inbox` case non-actionable.
- A mailbox whose own name contains `/` round-trips ambiguously through the new encoding. Worth at least a comment.
- Gmail users' `list-mailboxes` output and `Message.mailbox` change from `All Mail` to `[Gmail]/All Mail`. I think that's the right call, but it's a visible contract change filed under **Fixed** with no note — please call it out in the CHANGELOG, and consider a minor rather than a patch bump. Only the `list-mailboxes` description was updated; the other tool descriptions and the docs still describe leaf names.
- CHANGELOG heading is dated `2026-08-22`; should be `2026-08-21`.

---

Happy to take this in stages if you'd rather — the fragment fix plus the `!mailbox` change and the `parseRows` tests would get it to mergeable, and the forensics/rename items could follow in a second PR. Since mocked tests structurally cannot catch this class of bug, anything touching generated AppleScript needs a check against a real Mail.app before merge; I'm glad to run that on Gmail, iCloud and On My Mac here, but I don't have an Exchange account to verify your original case against.

sweetrb (changes requested):
Thanks for the thorough writeup and the regression tests — the container-walk approach (`mailboxPathFragment` reconstructing a path from `container of`) is the right fix for the underlying ambiguity, and `resolveAppleMailboxPath`'s exact-path-wins / unique-leaf / ambiguous-refuse resolution order is exactly what #197 asked for.

The gap is consistency: `resolveMailbox` now returns a compound canonical path for nested mailboxes (e.g. `Archive/Inbox`), but several call sites that consume that resolved string still address the mailbox with the old flat AppleScript form — `mailbox "X"` / `mailbox "X" of account "Y"` — which matches against the leaf-only `name` property. That form can't resolve a compound path, so those call sites regress on exactly the nested-mailbox case #197 describes, once a caller uses the canonical path the fix now advertises (e.g. via `list-mailboxes`, whose description this PR updated to say "use the full path for nested mailboxes").

Confirmed by reading the diff'd source directly:

1. **`search-messages` / `list-messages`, explicit non-Gmail mailbox param** (`appleMailManager.ts:2367` and `:2989`): `set theMailbox to mailbox "${escapeForAppleScript(targetMailbox)}"` where `targetMailbox` can now be `"Archive/Inbox"`. Wrapped in `try`, so it degrades to `notSearched`/`timedOut` in diagnostics rather than throwing — but a caller who takes the path `list-mailboxes` reports and searches it gets nothing back.
2. **`getUnreadCount`** (`:4838-4840`): `return unread count of mailbox "${safeMailbox}"` with no `try` at all around the command — a nested-path call errors out rather than degrading.
3. **`deleteMailbox`** (`:4920`) and **`renameMailbox`** (`:4995-4996`, `:5006`): same flat `mailbox "X" of account "Y"` form against the resolved (possibly compound) name.
4. **`createRule`'s move-to-mailbox action** (`:5764-5765`): untouched by this PR, same flat form — a rule targeting a nested mailbox by path, or a leaf name that's ambiguous, hits the same collision class #197 reports.
5. **`reconEmitFromMessage`** (`:1313-1322`, used by the unlocated-id branches of `findMessageScript`/`moveMessageInternal`/the batch handler): still does `set _umbName to (name of _umb)` — leaf-only — so the audit/forensics trail from #155/#157 can name and compare against the wrong mailbox for an id whose location wasn't pre-tracked, on exactly the accounts this PR is about.

One more, lower-severity: `searchMessagesWithDiagnostics`'s no-`account` fan-out (`:2253`, and the equivalent in `listMessagesWithDiagnostics` near `:2913`) has no `try/catch` around the recursive per-account call. Every other per-account failure mode in that loop (timeout, oversized mailbox) is captured into `diagnostics` and the loop continues — but `resolveMailbox`'s new ambiguous-leaf throw isn't caught there, so it unwinds the whole call and discards messages already gathered from previously-processed accounts instead of behaving like the other partial-result paths.

None of this undermines the core fix — `list-mailboxes`, the Gmail-receiving-set search/list paths, and `moveMessageInternal`/`batchMoveMessages` (which already used the container-walk match) are solid, and the new `appleMailManager.mailboxPaths.test.ts` cases are good coverage for `resolveAppleMailboxPath` itself. It's specifically the handful of call sites above that consume a resolved path but never got the container-walk treatment. Given `getUnreadCount`/`deleteMailbox`/`renameMailbox`/`createRule` are pre-existing code this PR didn't touch, it'd also be reasonable to scope this PR to search/list/list-mailboxes (already done well) and file a fast follow-up for the rest — happy to merge on that basis if you'd rather split it than extend this PR further.

sweetrb:
Pushed `99fa25b`, addressing both review rounds:

**Blocker 1/2 (container walk raises -1728).** Rewrote `mailboxPathFragment` to whitelist `mailbox`/`container` classes instead of blacklisting a bare `account` terminator Mail never actually reports — every accessor (`container of`, `class of`) is now wrapped in its own `try` so a future Mail change degrades to the leaf name instead of failing outright.

**Consistency gaps (search/list explicit mailbox, get-unread-count, delete-mailbox, rename-mailbox, create-rule).** All now resolve a compound path via the same container-walk match `move-message`/`batch-move-messages` already used, via a new shared `mailboxLookupFragment` helper, instead of the flat `mailbox "X"` form.

**Blocker 3 (forensics throw after a successful move).** `reconEmitFromMessage` now emits the canonical path instead of the leaf, and the `sameMailbox` comparison closures use a new non-throwing `resolveMailboxSafe` — so an ambiguous leaf can no longer make a *successful* move get reported as a failure.

**Blocker 4 (rename-mailbox orphan).** `oldName` is now resolved — and refused, if ambiguous — *before* `createMailbox` runs, not after. Added a regression test that asserts no destination is created when `oldName` is ambiguous.

**Blocker 5 (rowsIncludeMailbox desync).** Fixed `mailbox === undefined` → `!mailbox` so it tracks the same falsy branch the row-emission code actually takes. Added a dedicated regression test file (`appleMailManager.rowsIncludeMailbox.test.ts`) covering the `mailbox: ""` case for both search and list, including that the cached location used for a later by-id read is the real mailbox, not the fallback.

**Smaller items:** the no-`account` search/list fan-out now catches a per-account ambiguous-leaf throw into the same partial-results diagnostics a timeout gets, instead of discarding messages already gathered from other accounts. CHANGELOG dated correctly, bumped to a **minor** (2.17.0) since Gmail's nested mailboxes now report their real path (`[Gmail]/All Mail`) rather than the leaf — a visible contract change — and README calls that out plus documents the full-path/unique-leaf/ambiguous-refuse rule everywhere `mailbox` is accepted.

704 pre-existing tests pass (5 needed updating — they asserted the old flat lookup / old buggy terminator condition), plus 5 new tests. Lint/typecheck/format/build all clean.

sweetrb (approved):
Verified `99fa25b` against both review rounds.

**Blockers 1/2 (container-walk `-1728`).** The new `mailboxPathFragment` is character-for-character the fragment I live-verified against Gmail and On My Mac in the first review round — whitelists `mailbox`/`container` classes instead of blacklisting the `account` terminator Mail never reports, and wraps every accessor in its own `try`. Every other call site (`mailboxLookupFragment`, the delete/rename/get-unread-count/create-rule paths) is a mechanical reuse of that same primitive, not new unverified AppleScript.

**Consistency gaps from round 2** — search/list explicit mailbox, `get-unread-count`, `delete-mailbox`, `rename-mailbox`, `create-rule`'s account-scoped `moveTo` — all now resolve compound paths via the container-walk match, matching what `move-message`/`batch-move-messages` already did. The unscoped (no `moveToAccount`) `create-rule` path is left on the flat form, which is fine per my round-2 note that a narrower scope + fast-follow was an acceptable split.

**Blocker 3 (forensics throw after a successful move)** — `resolveMailboxSafe` + canonical-path RECON records fix the false-failure-report case.

**Blocker 4 (rename-mailbox orphan)** — `oldName` now resolves (and refuses, if ambiguous) before `createMailbox` runs; nice touch adding rollback-on-`newName`-ambiguity too, which is beyond what I'd asked for.

**Blocker 5 (`rowsIncludeMailbox` desync)** — `!mailbox` fix plus the new dedicated `appleMailManager.rowsIncludeMailbox.test.ts` covering all four row shapes closes the gap the original 704-test suite missed entirely.

I couldn't get a fresh live Mail.app run in this pass — the console is locked and Apple Events to Mail/System Events are timing out as a result, not a permissions issue (TCC grants all check out). Given the core previously-hard-failing logic is byte-identical to what I already ran live against Gmail and On My Mac, and everything downstream of it is either a mechanical reuse of that verified primitive or pure TypeScript covered by new regression tests, I'm comfortable merging on that basis. 709/709 tests pass, lint/typecheck/format all clean, CI green including `imap-integration`.

Thanks for the very thorough follow-through across two rounds of fairly demanding review — this is exactly the level of rigor this class of bug needs.

sweetrb:
Merged — thanks for the sustained follow-through here, @Tal-Golan. Two rounds of fairly demanding review (the container-walk `-1728` that 704 mocked tests structurally couldn't see, then the call-site consistency sweep) and both landed clean: `mailboxPathFragment`/`mailboxLookupFragment` reused everywhere a mailbox gets addressed, the forensics false-failure fix, the rename-orphan rollback, and the `rowsIncludeMailbox` field-count desync with dedicated regression coverage the original suite missed. That's real rigor on a class of bug that's easy to ship broken and hard to catch. Shipped as 2.17.0.

PR #200, opened 2026-08-24
fix(imap): search all provider mailboxes

### LINKED ISSUE #199: Unscoped IMAP search assumes Gmail All Mail and misses iCloud folders
## Summary

`search-messages` documents that omitting `mailbox` searches all mailboxes, but the IMAP path does not do that for iCloud/generic IMAP:

- a search scoped to an explicit iCloud account resolves the missing mailbox to Gmail's hard-coded `[Gmail]/All Mail` and fails to select it;
- the no-account fan-out avoids that error by forcing every non-Gmail provider to `INBOX`, which silently misses messages filed anywhere else.

The bug is still present on `main` / v2.17.0 in `resolveMailboxPath()` and `fanOutImapMessages()`.

## Reproduction

1. Configure an iCloud IMAP account.
2. Put a matching message outside `INBOX` (for example in Archive or Deleted Messages).
3. Call:

```json
{
  "tool": "search-messages",
  "arguments": {
    "account": "iCloud",
    "subject": "known subject"
  }
}
```

## Actual

The explicit-account path tries to select `[Gmail]/All Mail` on iCloud and errors. The unscoped multi-account path searches iCloud's `INBOX` only and returns a confident false negative for matches in other folders.

## Expected

An omitted mailbox should search the provider's complete searchable mail store:

- use the server-advertised RFC 6154 `\\All` mailbox when available (for example Gmail);
- otherwise search every selectable mailbox and globally merge/sort/limit the matches;
- report individual mailbox failures as partial coverage instead of hiding them.

## Regression evidence

A failing unit guard reproduces the current single-`[Gmail]/All Mail` selection. With the proposed fix, a read-only live iCloud probe searched 115 selectable mailboxes, found the known message in `Deleted Messages`, and returned `partial: false` with no failed mailboxes.

PR to follow.

fkoehler:
Closes #199.

## What changed

- make an omitted IMAP search mailbox provider-neutral instead of resolving it to Gmail's hard-coded `[Gmail]/All Mail`;
- discover and use a server-advertised RFC 6154 `\\All` mailbox when available;
- otherwise search every selectable mailbox, skipping `\\Noselect` hierarchy containers;
- merge the newest candidates globally, de-duplicate copies by Message-ID, then apply offset/limit;
- surface per-mailbox failures through `partial`, `failedMailboxes`, and the existing merged-search coverage diagnostics;
- remove the hostname-based Gmail/non-Gmail fan-out workaround, since the server's advertised mailbox metadata is authoritative;
- bump the shipped/plugin version to 2.17.1 and rebuild the committed bundle.

## Regression guard

The new guard was run before the implementation and failed in three places: the provider-neutral fallback was still `[Gmail]/All Mail`, an iCloud-like server selected only that path instead of its selectable mailboxes, and a failed folder was not reported as partial coverage.

After the implementation:

- `vitest run`: 50 files, 710 tests passed;
- focused IMAP + output-schema tests: 146 passed;
- TypeScript typecheck: passed;
- ESLint: passed with the 10 existing warnings, 0 errors;
- Prettier check: passed;
- committed build and plugin-version sync checks: passed.

## Live iCloud verification

A read-only search against a configured iCloud account searched 115 selectable mailboxes, found the known message in `Deleted Messages`, and returned `partial: false` with no failed mailboxes. The full 115-folder scan took about 75 seconds on that account; the implementation deliberately retains the project's single pooled IMAP connection instead of multiplying connection usage. No message mutation was performed.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.0",
+      "version": "2.17.1",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.0",
+      "version": "2.17.1",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.17.0",
+  "version": "2.17.1",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.17.0",
+  "version": "2.17.1",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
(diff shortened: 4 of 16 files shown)


sweetrb:
Merged as 2.17.1 — thank you, Fabian. This is a genuinely careful piece of work: the RFC 6154 `\All`-mailbox discovery with a full selectable-mailbox fallback, the global newest-first merge with Message-ID de-duplication across mailboxes, and reporting partial coverage via `failedMailboxes` instead of silently hiding a failed folder are all exactly right. The live 115-mailbox iCloud verification finding the message in Deleted Messages — and being explicit about the ~75s cost of that scan against the project's single pooled IMAP connection — made this easy to trust and merge with confidence.

I added two small follow-up commits on top of yours before merging: a README correction (the "Default mailbox is resolved per account" section still described the old Gmail-vs-iCloud heuristic your PR removes) and a CHANGELOG fix (the release-notes CI gate wants the [Unreleased] entry filed under a `[2.17.1]` heading once the version is bumped). Wanted to flag both so future PRs from you know what that gate expects.

PR #206, opened 2026-08-30
fix(smart-mailbox): use PlistBuddy Merge instead of crashing plutil -insert

### LINKED ISSUE #205: create-smart-mailbox always fails on macOS 15: `plutil -insert` crashes on a bare numeric keypath at an array root
## Summary

`createSmartMailboxAtPath` appends the new entry with

```
plutil -insert <index> -json <entry> <temp-plist>
```

On macOS 15 that call **crashes** instead of appending, so every
`create-smart-mailbox` fails. `delete-smart-mailbox` looks broken too, but only
because its test creates the entry first — its own PlistBuddy path is fine.

The failure is safe: `commitSmartPlist` rejects the copy and the original plist is
left untouched. Nothing is lost, the feature is just unusable.

## Reproduction — outside this project, 2 commands

```bash
printf '<?xml version="1.0" encoding="UTF-8"?>\n<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">\n<plist version="1.0"><array><dict><key>a</key><string>x</string></dict></array></plist>\n' > pt.plist

plutil -insert 1 -json '{"b":"y"}' pt.plist
```

```
*** Terminating app due to uncaught exception 'NSRangeException', reason:
'*** -[NSTaggedPointerString substringToIndex:]: Index 18446744073709551615
 out of bounds; string length 1'
...
3   plutil    __setKeypath_block_invoke + 196
4   plutil    processPlist + 1484
5   plutil    -[PLUContext execute] + 1224
```

The same tool with a **named** keypath on a **dict** root works:

```bash
plutil -insert arr.0 -json '{"b":"y"}' dict-rooted.plist   # exit 0
```

So it is specifically a bare numeric keypath at an **array** root — which is exactly
the shape `SyncedSmartMailboxes.plist` has.

## Environment

- macOS 15.7.7, Apple silicon
- apple-mail-mcp 2.17.2

## Test symptoms

`src/services/appleMailManager.smartMailbox.test.ts` — 3 failures on a clean
checkout:

```
× createSmartMailboxAtPath appends without touching the existing date/data entry
× createSmartMailboxAtPath is idempotent on an existing name
× deleteSmartMailboxAtPath removes only the named entry and preserves the rest
```

each with `{ created: false, error: "*** Terminating app due to uncaught exception 'NSRangeException' ..." }`.

Since these tests shell out to the real `plutil`, this should reproduce on any
macOS 15 CI runner — the tests are not at fault, they are correctly reporting a
broken call.

## Suggested fix

`PlistBuddy`'s `Merge` appends an array's items to an array root and preserves the
nested structure, including the `date`/`data` values that `plutil -convert json`
cannot represent. `deleteSmartMailboxAtPath` already uses PlistBuddy, so this would
keep both halves of the feature on one tool.

PR follows.

Zome59:
Fixes #205.

## Why

`plutil -insert <index> -json ... <plist>` crashes on macOS 15 when the plist has
an **array** root and the keypath is a bare number:

```
plutil -insert 1 -json '{"b":"y"}' array-rooted.plist
*** NSRangeException: -[NSTaggedPointerString substringToIndex:]:
    Index 18446744073709551615 out of bounds; string length 1
```

Reproducible outside the project — the two commands are in #205. The same call
with a named keypath on a dict root works, so it is the bare numeric keypath at an
array root that breaks, and that is exactly the shape of
`SyncedSmartMailboxes.plist`.

Effect: every `create-smart-mailbox` failed. It failed *safely* — `commitSmartPlist`
discarded the edited copy and the original plist was left untouched — but the
feature was unusable, and `delete-smart-mailbox` looked broken too because its test
creates the entry first. Its own PlistBuddy path was never at fault.

## How

`PlistBuddy`'s `Merge` appends an array's items to an array root and preserves
nested structure, including the `date`/`data` values the module's own comments note
that `plutil -convert json` rejects. The entry is written as a one-element array to
a temp JSON file, converted to `xml1`, merged, and both temp files are removed on
every path — including the two early-return branches.

`deleteSmartMailboxAtPath` already uses PlistBuddy, so both halves of the feature
now sit on one tool.

## Testing

- The 3 previously failing `appleMailManager.smartMailbox.test.ts` cases pass.
- Full suite **714/714** (was 711/714 on a clean checkout on macOS 15).
- `tsc --noEmit` clean; `build/index.js` regenerated as the repo convention expects.

No new dependency, no API change. The only behavior change beyond "it works" is the
fallback error string, from `plutil insert failed` to `PlistBuddy merge failed`,
plus a distinct message if the entry cannot be encoded at all.

## Note

I did not add a regression test — the existing cases already pin this and were red
on a clean checkout for exactly this reason. Happy to add one that asserts the crash
signature if you would rather have it explicit.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.2",
+      "version": "2.17.3",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.2",
+      "version": "2.17.3",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.17.2",
+  "version": "2.17.3",
   "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
   "author": {
     "name": "Rob Sweet",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "apple-mail-mcp",
-  "version": "2.17.2",
+  "version": "2.17.3",
   "description": "Apple Mail integration for Claude Code and Codex via MCP",
   "owner": {
     "name": "Rob Sweet",
@@ -12,7 +12,7 @@
       "name": "apple-mail",
       "displayName": "Apple Mail",
       "description": "Manage Apple Mail through natural language - read, search, send, and organize emails (macOS only)",
-      "version": "2.17.2",
+      "version": "2.17.3",
       "author": {
         "name": "Rob Sweet",
         "email": "rob@superiortech.io"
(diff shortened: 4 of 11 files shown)


sweetrb:
Thanks for this — clean fix, and the PlistBuddy `Merge` approach is exactly right (keeps both smart-mailbox code paths on one tool, and it's what already handles the `date`/`data` values `plutil -convert json` can't). Merged as 2.17.3, with a version bump + CHANGELOG entry folded in (this repo's release convention requires shipped-code changes to bump the version in the same PR — added that on your branch since maintainer edits were enabled). 714/714 tests green.

Issue #12, opened 2026-05-13
send-email: outgoing AppleScript-built messages get wrapped in <blockquote type="cite"> (URLShareWrapperClass) on macOS 15+


sweetrb:
## Symptom

Emails sent via `mcp__apple-mail__send-email` render to recipients as if they were forwarded/quoted messages.

In the MIME source:
- **text/plain part**: every line is prefixed with `> `, signature appears below outside the quoted section.
- **text/html part**: body is wrapped in `<blockquote type="cite">` under `<div class="Apple-Mail-URLShareWrapperClass">` (and preceded by `<div class="Apple-Mail-URLShareUserContentTopClass">`). Signature is appended as a sibling `<div>` outside the blockquote.

Confirmed on:
- macOS 26.4.1 ("Tahoe"), Mail.app 16.0 build 3864.500.181 and 3864.600.51.1.1

## Reproducer

Any of these produce the same bad output:

```typescript
// Pattern A — current sendEmail
set newMessage to make new outgoing message with properties {
  subject: "...", content: "...", visible: true
}

// Pattern B — set content after creation, inside tell block
set newMessage to make new outgoing message with properties {
  subject: "...", visible: false
}
tell newMessage
  set content to "..."
  -- recipients, sender, send
end tell

// Pattern C — Pattern B + default message format = plain format
tell application "Mail"
  set default message format to plain format
  -- pattern B body
end tell
```

All three produce identical `Apple-Mail-URLShareWrapperClass` + `<blockquote type="cite">` wrapping.

## Upstream root cause

Apple Mail's AppleScript content insertion path on macOS 15+ routes through the same compose template used by the Share API, which wraps the body. Apple radar **FB11734014** is open since Ventura with no fix; users have been hitting this since 2022.

Related external discussion:
- https://forums.macrumors.com/threads/applescript-creating-a-new-message-in-mail-app-is-causing-weird-formatting-issues.2385052/
- https://developer.apple.com/forums/thread/738842

## Possible fixes

Listed from least to most invasive:

1. **Document the limitation and offer guidance.** Most users of this MCP won't notice (the email IS delivered and readable). For users who do (e.g. `rob@superiortech.io`'s NHL bracket digest, where the visual artifact is jarring), point them at the SMTP-direct workaround.
[...]

PR #214, opened 2026-09-04
fix: preserve clean SMTP delivery for IMAP replies

jjoanna2-debug:
## Fix

Composite imap: IDs were passed to a numeric AppleScript source lookup, silently returning SMTP replies to the body-wrapping path in #12.

Read originals from their IMAP account/mailbox/UID with a 25 MiB bound. Preserve reply headers and unquoted new text. Add explicit SMTP/AppleScript selection, account-identity checks, and no silent fallback after selecting SMTP. Drafts retain AppleScript. SMTP forwards without a readable plain-text original now fail before sending; HTML-only originals require explicit AppleScript forwarding. Version 2.17.8, docs, manifests, skills and committed bundle are synchronized.

## Verification

748 unit tests across 52 files; lint, typecheck, formatting, eight built-server schema tests, and deterministic double build passed. Regression tests fail when IMAP routing or In-Reply-To is removed. Real Nodemailer MIME tested without delivery. Additional regression tests reproduced silent body omission for HTML-only IMAP messages and failed Mail.app body reads on the original head; both pass with the correction.

Original-head run: https://github.com/jjoanna2-debug/apple-mail-mcp/actions/runs/33915613764

Correction: 73a8a417401bcf353a9b6f12f7c6d130ece0f02d. Local lint (zero errors, ten existing warnings), typecheck, formatting, all unit tests, eight built-server schema tests, skill/version synchronization, and deterministic rebuild passed. Current-head [CI](https://github.com/sweetrb/apple-mail-mcp/actions/runs/33937852710), [CodeQL](https://github.com/sweetrb/apple-mail-mcp/actions/runs/33937852722), and [version guard](https://github.com/sweetrb/apple-mail-mcp/actions/runs/33937852651) passed.

Two commits on current upstream main; no workflow changes. No live configured Mail.app or provider SMTP delivery was run. Selected-SMTP failures now return errors rather than falling back; this compatibility change is documented. Apple's underlying compose bug and local SMTP Sent-copy behavior are unchanged.

Related: #12.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -6,7 +6,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.7",
+      "version": "2.17.8",
       "source": {
         "source": "local",
         "path": "./codex"
--- a/.antigravity-plugin/marketplace.json
+++ b/.antigravity-plugin/marketplace.json
@@ -7,7 +7,7 @@
   "plugins": [
     {
       "name": "apple-mail",
-      "version": "2.17.7",
+      "version": "2.17.8",
       "description": "Manage Apple Mail through natural language - read, search, send, reply, forward, and organize emails and mailboxes, with diagnostics, attachments, templates, and mail-merge support (macOS only).",
       "source": {
         "source": "local",
--- a/.antigravity-plugin/plugin.json
+++ b/.antigravity-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "apple-mail",
-  "version": "2.17.7",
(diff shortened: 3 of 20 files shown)


sweetrb:
Merged — thanks for tracking this down. The composite `imap:` id being handed to the numeric AppleScript lookup (silently falling into the body-wrapping path from #12) was a subtle one, and reading the source directly over IMAP with the account-identity check is the right fix — it also closes a latent cross-account-send gap that fallback path never guarded against. Appreciated the thorough regression coverage on the failure-no-fallback behavior and the rendered-MIME test proving the thread headers actually reach the wire.

Issue #229, opened 2026-09-13
Non-RFC-5322 (Spanish locale) Date header still merges into Subject in v2.19.2 — reproduced with raw IMAP bytes


j5pu:
## Summary

Following up on #226 (closed) / #228 (merged, v2.19.2): the header-fusion bug
that #226 originally reported is still present in v2.19.2, for messages whose
raw `Date:` header uses a non-RFC-5322 Spanish-locale format (e.g. `mié oct 10
14:25:15 2007`, from legacy Microsoft Entourage / Outlook for Mac). Confirmed
with a fresh install via `@latest` (`npm view apple-mail-mcp version` → 2.19.2,
no newer version published), and confirmed against the raw IMAP source to rule
out corrupted/empty headers as the cause.

## What #228 fixed (and didn't)

#228 correctly separated `dateSent`/`dateReceived` for well-formed headers,
and #227 fixed a different bug (CR-only line breaks dropping all headers).
Neither addressed the original non-RFC-5322 Spanish-locale date parsing
failure reported in #226.

## Reproduction

Account: iCloud, mailbox "XXXX".

Message A — id 345559 (also 339973), subject "diferencial":
`get-message-headers` returns:
```
"Date": "Subject: diferencial"
```
with `Subject` entirely absent from the parsed headers, and
`dateReceived: "2014-01-14T12:53:59.000Z"` (a known corrupted-migration
INTERNALDATE; real year is 2007-2008).

Message B — id 1189559/1189561, subject "RE: puedes leer este
www.XXXXX.XXX/articulos/XXXXXX.htm": same fusion pattern —
```
"Date": "Subject: RE: puedes leer este www.XXXXX.XXX/articulos/XXXXXX.htm"
```

Message C — id 1186096, subject "RE: ", from a different known
corrupted-migration cluster (dateReceived 2022-08-05): same fusion —
```
"Date": "Subject: RE:"
```
`get-message` on this id returns `dateSent: "2025-09-07T05:09:41.000Z"` — a
value with no relation to the message (looks like a parse-failure fallback,
not the real header date or an explicit null/error).

## Ground truth (raw IMAP, bypassing the connector entirely)

For message C (Message-Id `<337BB11B-4C26-4786-B301-AF7FBAD16902@XXXXX.XXX>`),
fetching `BODY.PEEK[HEADER]` directly via `imaplib` against the same
account/mailbox gives:

```
[...]

Issue #226, opened 2026-09-12
get-message-headers / dateSent mishandles non-RFC-5322 Date: headers (drops next header, e.g. Subject)


j5pu:
Follow-up to #223/#224/#225 — thank you for the dateSent/get-message-headers fix, it works great for normal messages. However, it doesn't help for the exact migrated-mailbox use case that motivated it, because a large subset of the old messages have a non-RFC-5322-compliant Date: header and the parser mishandles that case.

**Root cause**: messages sent by old Microsoft Entourage / Outlook-for-Mac (~2005-2011, identifiable by MIME boundary OUTLOOK2MACxxxxxxxx) wrote the Date: header using the OS locale instead of RFC 5322 format, e.g. (Spanish locale):

```
Date: jue ago 30 13:55:12 2007
Subject: diferencial
```

("jue" = Thursday, "ago" = August, in Spanish — not English, and not the required `Thu, 30 Aug 2007 13:55:12 +0200` shape.)

**Observed bug**: when Date: fails to parse, the header parser appears to consume the next header line into it instead of preserving the raw text and moving on. get-message-headers on this exact message returns:

```json
{"dateHeader": "Subject: diferencial", "headerCount": 6, ...}
```

Note Subject is missing entirely from the headers array/subject field — its line got merged into the Date value. Confirmed reproducible on a second independent message from the same mailbox/era (same boundary marker, different subject) with the identical failure signature.

**Minimal repro** (raw header block):
```
From: Test Sender <test@example.com>
To: 'Test Recipient'
Date: jue ago 30 13:55:12 2007
Subject: diferencial
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="OUTLOOK2MAC8473928"
```

**Expected**: dateHeader = "jue ago 30 13:55:12 2007" (raw, even if unparseable), subject = "diferencial", date (ISO) absent/null if truly unparseable — but no header should vanish or get merged.

**Suggested fix**: (1) never let a failed date-parse consume the following header line — treat header-line splitting independently of value validation; (2) optionally add a tolerant fallback parser for localized weekday/month abbreviations (at minimum Spanish: lun/mar/mié/jue/vie/sáb/dom, ene/feb/mar/abr/may/jun/jul/ago/sep/oct/nov/dic) since this exact pattern — old Entourage/Outlook-for-Mac mail in a non-English locale — is presumably common in exactly the kind of legacy/migrated mailboxes this feature targets.

[...]

PR #231, opened 2026-09-13
fix: tolerate non-English month abbreviations in Date: headers

sweetrb:
Closes the reproducible half of #229, using the byte-exact header block @j5pu supplied.

## Root cause

Legacy Entourage / Outlook for Mac wrote the **system locale's** month abbreviation instead of RFC 5322's English one. A mailbox migrated from them carries:

```
Date: jue ago 30 13:55:12 2007
```

`new Date("jue ago 30 13:55:12 2007")` is **NaN** — `ago` is *agosto*, not `Aug` — so the header date was lost for every affected message.

## Why it went unnoticed

⚠️ **The failure is partial.** Spanish `oct` is identical to English `Oct` and parses fine; `ago`, `ene`, `abr` and `dic` do not. A spot check on an October message reports the mailbox healthy — and that is exactly what misled my own first analysis of this issue, which used `oct` and concluded the locale format parsed correctly.

```
new Date("jue oct 30 13:55:12 2007") -> 2007-10-30   ← coincidence
new Date("jue ago 30 13:55:12 2007") -> NaN
new Date("jue ene 30 13:55:12 2007") -> NaN
new Date("jue abr 30 13:55:12 2007") -> NaN
new Date("jue dic 30 13:55:12 2007") -> NaN
```

## The fix

`parseDateHeader` tries `Date.parse` first, then maps a recognised es/fr/de/it/pt month abbreviation and retries. A genuinely unusable value still yields `undefined` — omitted, never invented, matching the IMAP rows (2.19.2) and the AppleScript path (2.19.3). `dateHeader` continues to carry the verbatim string either way, so an unparsed value stays visible.

Only the **month** is mapped; the weekday is decorative and V8 ignores an unrecognised leading token. Word-boundary anchored so it cannot rewrite part of a timezone name.

## Tests

Four, all confirmed to **fail with the fix reverted**, built on the reporter's byte-exact block (From/To/Cc masked at his request; Date/Subject untouched). One deliberately covers the months that silently worked alongside the ones that did not, so the partial nature of the failure is pinned.

**Two pre-existing tests changed, and I want to be explicit about it.** `never drops or merges a header just because Date: is an unparseable locale string (LF/CRLF)` asserted `date === undefined` for this exact string. Those tests' subject is header *integrity*; the `undefined` assertion recorded the limitation this change removes. They now assert the parsed value — every integrity assertion in them is untouched.

## What is still open on #229

The report also describes `Date:` absorbing the next header line, with `Subject` vanishing. **That does not reproduce from the supplied bytes**: the block is clean CRLF, no leading whitespace, no non-ASCII, and `parseHeaderBlock` returns `Date` and `Subject` as separate fields. A regression test covering exactly that block has existed since #226. Since the supplied artifact is from the IMAP path, the remaining suspect is Mail's `all headers` on the AppleScript path — asked for on the issue.


j5pu:
Thanks. I tested and everything works now. Perfect.

PR #241, opened 2026-09-18
feat: add explicit draft identity and guarded saved-draft sending

maxschuetz06:
## Description

When an agent creates a draft using a Mail account label, `create-draft` currently passes that label directly to the `sender` property. In a multi-account setup this can silently select the wrong identity. The draft operation also reports success without explicitly saving or checking the result, and there is no native-signature selection or existing-draft send operation.

This contribution adds:

- Account-name/address resolution and an explicit `sender` override, including aliases. Missing, disabled, ambiguous, or mismatched identities fail before composing.
- `list-signatures` and an optional exact `signature` name on `create-draft`. Mail inserts its native signature; no global preference is changed.
- Explicit save and readback checks, with a structured receipt containing the actual sender, signature, and `composeId`. Compose mutations are never automatically retried.
- `send-saved-draft`, which checks a fresh saved Drafts locator and the original composer against the approved sender, recipient, subject, body, and signature. It defaults to `dryRun: true`; only `dryRun: false` submits mail. Content comparison ignores whitespace while preserving punctuation, case, and diacritics.

### Deliberate limits

`send-saved-draft` only supports MCP-created drafts whose original outgoing object still exists in the same Mail session. It does not reopen arbitrary drafts or recreate messages. This first implementation supports one To recipient, a named signature, and no CC/BCC or attachments. `composeId` is not a stored message ID. A submission receipt is not recipient delivery confirmation; callers must verify Sent and never blindly retry an uncertain result.

The existing immediate-send and SMTP paths are unchanged. Version bumped to 2.19.8 and plugin manifests, generated skill copies, changelog, and committed bundle synchronized per CONTRIBUTING.md. No local installation overrides, account details, or correspondence are included.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Refactoring — existing AppleScript escaping helpers extracted without changing behavior

## Testing

- [x] Full unit suite with coverage on Node 24: **864 tests passed** across 60 files; coverage gates pass.
- [x] Built-server MCP output-schema contract: **8 tests passed**.
- [x] Lint passes: no errors, 10 existing warnings.
- [x] Typecheck, formatting, build, plugin-version and generated-skill checks pass.
- [x] Native macOS tests compile both preview/send scripts without executing Mail commands, and exercise whitespace/punctuation comparison without accessing Mail.
- [x] Sender/signature selection, persisted draft content, and existing-composer sending were exercised through the local stdio MCP against Mail.app; saved and Sent copies were independently read back. No additional emails were sent while preparing this PR. The final stricter comparison has dedicated native regression coverage.

Browser QA is not applicable: this changes a native Mail/MCP backend, not browser UI. The entire general live-Mail integration suite was not run against the personal mailbox. An older local Node 22.15 runtime reproduces the pre-existing Windows-1252 Euro-decoding failure on the unchanged base; the complete suite passes on Node 24.

## Checklist

- [x] I have read CONTRIBUTING.md.
- [x] Code follows the project's style.
- [x] Documentation updated.
- [x] Patch version bump, changelog entry, synchronized manifests and rebuilt bundle included.

Opening as a draft for feedback on the API and the explicitly limited saved-draft workflow. Thank you for maintaining this project.


sweetrb (changes requested):
Thanks for opening this as a draft and asking for API feedback first, @maxschuetz06 — that's the right call for a change that adds a send path. I read the full diff (source, tests, docs; not the bundle line-by-line). Verdict: **the direction is good, but it isn't ready to un-draft yet** — there are three design questions that gate the shape of `send-saved-draft`, and four things I'm fairly sure are defects on real accounts.

### What's right

- The `set sender to "<account label>"` bug is real. Handing Mail an account *name* as `sender` doesn't match any address, so Mail silently falls back to the default account. Resolving name/address → an enabled account's actual address *before* `make new outgoing message`, and failing on missing/disabled/ambiguous, is exactly the fix. (The immediate-send path in `sendEmailWithPaths` has the same `set sender` line — worth noting in the PR body that this fix is draft-only for now.)
- Explicit `save` + readback receipt, and `maxRetries: 1` on every compose mutation. Correct: a timed-out compose may have already produced a draft, so retrying is how you get duplicates. Your point also surfaces that the existing `send-email` AppleScript path still runs with `maxRetries: 2`, and `executeAppleScript` retries on timeout — I'll handle that one separately, it's not this PR's job.
- Housekeeping is all there: patch bump, CHANGELOG under a real `## [2.19.8]` heading with `[Unreleased]` left empty, all six manifests + three skill copies synced, bundle rebuilt. That's what `require-version-bump` and the skill-drift check gate on.
- Structured `Use when / Returns / Do not use when / Safety:` descriptions on both new tools, and `dryRun` defaulting to `true`.

### Design questions (these decide the API, please answer before un-drafting)

**1. Does the composer survive `create-draft`?** `buildDraftScript` ends with `close newMessage saving yes`. `send-saved-draft` then requires `every outgoing message whose id is <composeId>` to return exactly one. In Mail's object model `outgoing messages` are the open composers, and `close` disposes one — so as written I'd expect every `send-saved-draft` after a `create-draft` to fail with "Original composer unavailable". The PR body says the composer send was exercised live, so I may be wrong about what `close saving yes` does to an invisible outgoing message. Could you post the exact sequence you ran — `create-draft` result, the `list-messages` on Drafts that produced `draftId`, and the `send-saved-draft` call with its output? If the object does survive, please say what keeps it alive; if it only worked without the `close`, the design needs a different anchor than `composeId`.

**2. What is the review surface?** The stored-body comparison fails closed on *any* edit, so the user cannot touch the draft in Mail.app between `create-draft` and `send-saved-draft`. That means approval is happening in chat — and for chat-approved sends `send-email` already exists. What's left as the value of the round-trip is: native signature inserted by Mail, and a verified From identity. If that's the goal, say so in the descriptions and drop "review in Mail.app" from `create-draft`'s. If instead the goal is "the user may edit in Mail, then the agent sends what's actually in Drafts", that's a different (and simpler) tool: send the stored draft after echoing its current content back for confirmation. Which one do you want?

**3. Mail.app-only by construction.** Numeric `draftId`, AppleScript throughout — fine for a composer-based feature, but this server routes reads to IMAP whenever an `APPLE_MAIL_MCP_IMAP_*` account is configured, so most users get `imap:` ids from `list-messages` and will hit the `^\d+$` regex with no hint why. The `draftId` description should say "numeric Mail.app id from `list-messages` on Drafts with `transport: applescript`" (or whatever the right incantation is), not just "fresh stored Drafts id".

### Likely defects

- **Sender comparison in `draftSend.ts`.** `sender of sourceMessage is not "<addr>"` and the same for `targetMessage`. Mail returns `Full Name <addr>` whenever the account has a name set, which is nearly always. `buildDraftScript` already tolerates that (`does not end with ("<" & selectedSender & ">")`); the send script doesn't, so any named account fails "Stored draft sender mismatch".
- **`mailbox "Drafts" of sourceAccount`.** Gmail accounts in Mail.app keep drafts under `[Gmail]/Drafts`, a nested mailbox, so this errors for Gmail (and any provider that names it differently — see the `MAILBOX_ALIASES` table in `appleMailManager.ts`, which exists for exactly this). Mail's `account` class has a `drafts mailbox` property that resolves the special mailbox for every provider; use that.
- **`signature` is required on `send-saved-draft`** (`z.string().min(1)`, plus `error "Missing signature"`), but optional on `create-draft`. A draft created without a signature can never be sent by this tool. Make it optional and, when omitted, verify the composer has *no* signature.
- **`account` on `send-saved-draft` is name-only** (`account "<name>"`), while `create-draft` accepts name or address. Same parameter, two contracts. Also: AppleScript string comparison is case-insensitive unless inside `considering case`, so the docs' "exact" account/signature matching is really case-folded — either say so or wrap those checks too.

### Consistency nits

- `send-saved-draft` imports `sendSavedDraft` from the service directly; every other tool goes through `mailManager`. Add a `mailManager.sendSavedDraft()` passthrough like you did for `listSignatures` so mocks and any future router seam live in one place.
- `ASCII character 31` → `character id 31`. `ASCII character` is a Standard Additions command that costs an Apple Event per call; the notes server just swept these out for a measurable win.
- `utils/appleScriptText.ts` next to `utils/applescript.ts` is a confusing pair. Either fold the escape helpers into the existing module or name the new one for what it holds. The re-export from `appleMailManager.ts` for compatibility is fine.
- `osacompile` of a `tell application "Mail"` block inside a unit test is fine on `macos-latest`, just confirm it resolves terminology without launching Mail on the runner.
- README: `send-saved-draft` needs the parameter table the other tools have (with `draftId`'s type and the numeric-id caveat), and the `**⚠️ Safety:**` callout the other send tools carry. `CLAUDE.md`'s "send-email vs create-draft" section should mention the new path.

Once the three questions are settled I'm happy to help finish it — `maintainerCanModify` is on, so the remaining bump/doc fixups can land on your branch directly.

sweetrb:
Follow-up to the review: the retry issue your PR surfaced on the *existing* send/draft paths is now fixed on `main` as #242 (`maxRetries` 2 → 1 on `sendEmailWithPaths` and `createDraftWithCommands`, with a test pinning it). Two practical consequences for this branch:

- **2.19.8 is taken** — when you rebase, `pnpm version patch --no-git-tag-version` again so this lands as 2.19.9, and move your CHANGELOG entry under that heading.
- `createDraftWithCommands` now carries a comment + `maxRetries: 1` on the hunk your PR deletes, so expect a trivial conflict there; keeping your replacement is the right resolution.

PR #243, opened 2026-09-20
feat(search): body parameter for server-side IMAP body search

isking-byte:
## Description

Adds a `body` parameter to `search-messages` so a message can be found by words in its body, not only its subject or sender.

On an IMAP-configured account it adds an IMAP `BODY` criterion, combined with the other filters. I found I needed it when looking for an email where the name I was searching for only appeared in the text. `query` only matches subject and sender, so the search came back empty and the only way to find it was to list everything from that sender and read each one.

`query`'s schema description said "subject, sender, or content", which didn't match what either backend does, so I've corrected it to "subject or sender".

Body search is IMAP-only, because AppleScript's `content contains` has to pull every body through the Apple Event bridge. Rather than drop the filter quietly and return subject or sender matches as though they were body matches:

- an explicit AppleScript account returns an error explaining IMAP is needed
- an unscoped search lists AppleScript-only accounts under `notSearchedMailboxes`

I've tested it against iCloud (`imap.mail.me.com`), which supports `SEARCH BODY`, on a mailbox of around 2,200 messages.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Testing

- [x] Tests pass locally (`pnpm test`)
- [x] Linting passes (`pnpm run lint`)
- [x] Typecheck and formatting pass (`pnpm run typecheck && pnpm run format:check`)
- [x] Build succeeds (`pnpm run build`)

Two unit tests cover the new criterion, on its own and combined with `query`, `from` and a date. I haven't run the integration suite. The only AppleScript-path change is the early error when `body` is used without IMAP.

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) guidelines
- [x] My code follows the project's style
- [x] I have updated documentation if needed
- [x] Version bumped to 2.19.9 with a CHANGELOG.md entry


sweetrb:
Thanks for this, @isking-byte — the design is right on every count: keeping `body` as an explicit, additive IMAP `BODY` criterion rather than silently folding it into `query`, refusing to fake a body match on AppleScript-only accounts (and surfacing them via `notSearchedMailboxes` instead of a silent skip), and catching that the `query` schema description overclaimed content matching it never did. Tests for the criterion alone and combined with the other filters, README/CHANGELOG/version bump/plugin manifests all landed together — this merged clean. Shipped as 2.19.9.

